PR #2, opened 2020-03-21
Keep example formatting

cgreene:
Keep the example formatting around.


cgreene:
I'm not sure why this had travis enabled since I don't think it was ever intentionally turned on. I've now, hopefully, turned it off and we'll only see github actions builds 🤞

PR #5, opened 2020-03-21
Added instructions for adding content

rando2:
Adapted from Casey's email to Sam. Assuming minimal familiarity with git and GitHub here -- we may want to add instructions for command line users as well?


Instructions.md
```diff
@@ -0,0 +1,81 @@
...
+3. Make any desired changes.
+4. Scroll down to the bottom. 
+There you will see a section that says "Commit Changes." 
+Give your submission a title in the top box and briefly summarize your changes in the bottom box. 
+5. Click the box that says "Create a new branch for this commit and start a pull request." 
+This will submit a request to add your changes to the underlying document and will notify us to integrate your text into the document!
+6. Don't forget to add your information to the [author list](https://github.com/greenelab/covid19-review/blob/master/content/metadata.yaml). 
+Copy and update this template:
+`-
+    github: your_username
```
cgreene: I think this might work better as a multi-line code block with ```

rando2: TIL! It looks way better!

PR #14, opened 2020-03-22
Readme Enhancements

mprobson:
- Add buttons to Open and Issue and Fork
- Fix broken Internal links
- Add internal links to sections
- Small whitespace edits

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ Please feel encouraged to ask questions by opening a [Request for Help issue](is
 This project is a collaborative effort that will benefit from the expertise of scientists across a wide range of disciplines!
 
 For git users, to open a pull request please:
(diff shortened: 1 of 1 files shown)


rando2:
Thanks for these fixes, @mprobson -- my only suggestion is to [add yourself at the bottom for software contributions](https://github.com/greenelab/covid19-review/blob/master/content/metadata.yaml). There is a formal process for determining authorship and this way we can track who contributed even if you end up meeting acknowledgements rather than authorship criteria.

mprobson:
You're welcome! To confirm, that should be submitted as a separate PR, correct? I'm 99% sure that's the case according to the instructions but figured I'd leave a trail here for future external contributors.

rando2:
Yes, please! [For now it seems it needs to trigger the manual action of me explicitly granting you read permissions.](https://github.com/greenelab/covid19-review/issues/15)

mprobson:
Added, please see #16

Issue #13, opened 2020-03-22
Strategies for Systematic Review


juliettemarie0405:
Hi all,

So excited that you are starting this initiative and very grateful to be involved! I was involved in a systematic review a couple years ago. Based on that experience, I have some thoughts/questions outlined below.

1. **Centralized database.** In order to synthesize the data we garner from the literature, wouldn't it be helpful to have information on each paper stored in a data format that is able to be searched and sorted? Is it possible to make a collaborative excel sheet on GitHub? That way we could list the paper information, name of the individual reviewing, whether it contains information relevant to diagnosis/therapeutics/other, the area of study (virology vs. epidemiology etc.), the summary of findings, and whatever other information we deem important. We would then have a centralized database that can be sorted and searched for relevant information when writing up the paper.

2. **Inclusion and Exclusion Criteria**. The typical systematic review process (from my experience) starts with an initial scan of titles and abstracts based on pre-specified inclusion and exclusion criteria. Those deemed relevant in that first pass are then subjected to a full-text scan based on another set of pre-specified inclusion and exclusion criteria. The resulting set of relevant papers are then subjected to data extraction. I think it might be important to get area experts involved in designing these inclusion and exclusion criteria. The literature at hand encompasses so many different topics that we would need help identifying those relevant to therapeutics and diagnostics (for example, do we want to include papers that estimate the reproduction number? or model the global spread of the pandemic? or are these questions beyond the scope of therapeutics/diagnostics?) Once we have these inclusion/exclusion criteria in place, we could crowd-source for help in the first and second pass scans of the literature.

In my experience, these scans of the literature are always done by hand, usually by two independent reviewers who talk through any discrepancies. When I was involved in this process, we used a software called [EROS](http://eros-systematic-review.org/) (but the link no longer seems to work...so it may not exist anymore?). Perhaps it would be possible to do this with machine learning? The main roadblock I can foresee is that abstracts and full-texts will include references to previous literature, so it would be difficult for a non-human to identify what the **primary data** from each paper is.

[...]

PR #25, opened 2020-03-23
Attempt Regexp Labeling

cgreene:
This was discussed in #13 and needs a fair bit of work first but I wanted to test something simple to see how it went and also to test case sensitivity.


cgreene:
For probably obvious reasons the new action doesn't run in the PR branch. I'm not sure how to test this other than live 😬 . Here goes nothing...

PR #29, opened 2020-03-23
Update metadata.yaml

nilswellhausen:


content/metadata.yaml
```diff
@@ -53,3 +53,16 @@ authors:
     coi:
       string: "None"
       lastapproved: !!str 2020-03-23
+    github: nilswellhausen
```
rando2: My comment is that I think you're missing the leading "-" before your info is added that denotes a new author. So it should look like:
```
  -
    github: nilswellhausen
```
If you can make that change, I will approve!

content/metadata.yaml
```diff
@@ -53,3 +53,16 @@ authors:
...
+    name: Nils Wellhausen
+    initials: NW
+    orcid: 0000-0001-8955-7582
+    email: nilsw@pennmedicine.upenn.edu
+    contributions:
+      - Writing
+      - Visualization
```
rando2: Awesome that you want to do visualization! For now, just write the contributions you have submitted PRs for (so e.g., if you are proposing edits to the paper structure, then add yourself for one or both of the writing categories [here](https://casrai.org/credit/)).

content/metadata.yaml
```diff
@@ -53,3 +53,17 @@ authors:
     coi:
       string: "None"
       lastapproved: !!str 2020-03-23
+    -
```
cgreene:
```suggestion
  -
```
I think the spacing might matter

content/metadata.yaml
nilswellhausen: Thank you! Casey and I fixed it now

content/metadata.yaml
nilswellhausen: Hi rando2! I submitted a pull request to add a section about neutralizing antibodies in the treatment section. It could be either under antivirals or as its own section. Comments about that are much appreciated!

PR #39, opened 2020-03-24
Add pathogenesis section

rando2:
@nilswellhausen & @rays1987 are interested in developing a section on pathogenesis. I made it number 7 to try to keep some space on either side in case we add additional sections


rando2:
Sorry for asking you all for a review! I think it should be a quick one to check but from the intros, you all seem to have the relevant experience to tell me if something went horribly wrong :)

rando2:
Thanks @agitter !

PR #48, opened 2020-03-25
Updates to README, Instructions, USAGE, added Contributing

rando2:
Made suggestions in #18 and tried to explain file structure more explicitly per #32

PR #52, opened 2020-03-25
Fix broken readme links

agitter:
Follows up on #48.  The links are case-sensitive.


rando2:
So sorry I skipped this, thank you for handling it @agitter !

PR #82, opened 2020-03-26
✨ Metadata update

LucyMcGowan:


content/metadata.yaml
```diff
@@ -108,3 +108,17 @@ authors:
     coi:
       string: "None"
       lastapproved: !!str 2020-03-25
+ -
```
rando2:
```suggestion
  -
```

rando2: From the error messages, looks like failure on checks is due to a whitespace issue on line 111 but possibly also line 112 -- hoping this fixes it.

LucyMcGowan:
thank you @rando2!

PR #88, opened 2020-03-26
Use short DOI to fix broken reference

agitter:


agitter:
Corrected reference from artifact
image

PR #116, opened 2020-03-30
Additional direction for summary in new comment

### LINKED ISSUE #115: Paper Summaries as New Comments?
Very minor issue but just thought I'd post it now as small improvements in the paper review process early on could have compounding effects later.

**Is your feature request related to a problem? Please describe.**
The new paper templates say that paper summaries should be created as new comments. 
It seems like this isn't the current practice (I don't see this being done in the new paper issues that I scanned)

**Describe the solution you'd like**
If there is a big upside to having the summaries in new comments, edit the template directions to be more clear and forceful. See #116 as an example of what this could look like.
If there isn't a strong need, remove the directions recommending this from the templates.

rdvelazquez:
An example of more forceful direction for putting paper summaries in new comments. If this is something with a big upside. 

I only edited the "other" template. I can provide examples for the therapeutics and diagnostics templates if it's helpful.

Closes #115

### DIFF
--- a/.github/ISSUE_TEMPLATE/new-paper-template.md
+++ b/.github/ISSUE_TEMPLATE/new-paper-template.md
@@ -44,9 +44,13 @@ If it's primarily relevant to Diagnostics or Therapeutics, please go back and us
 - [ ] pharmacology
 - [ ] other:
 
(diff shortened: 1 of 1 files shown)


rando2:
Directions are very clear now, thank you for updating them!

Issue #109, opened 2020-03-29
New Paper (Other): Potential roles of social distancing in mitigating the spread of coronavirus disease 2019 (COVID-19) in South Korea


alavendelm:
Title: Potential roles of social distancing in mitigating the spread of coronavirus disease 2019 (COVID-19) in South Korea

# General Information
Link: https://github.com/parksw3/Korea-analysis/blob/master/v1/korea.pdf
Data & code link: https://github.com/parksw3/Korea-analysis 

## What is the paper's [Manubot-style citation](https://github.com/greenelab/covid19-review/blob/master/USAGE.md#citations)?

Citation:

### Is this paper primarily relevant to Background or Pathogenesis? 

- [x] Background
- [x] Pathogenesis
- [ ] Methods

## Please list some keywords (3-10) that help identify the relevance of this paper to COVID-19

* population dynamics 
* transmission 
* social distancing 
* contact tracing 
* reproductive number
* time-dependent reproductive number

- [ ] virology
- [x] epidemiology
- [x] biostatistics
- [ ] immunology
- [ ] pharmacology
- [ ] other:

# Summary
If you would like to submit a summary of the paper, please copy and paste the following into a comment.

### Suggested questions to answer about each paper:
- What did they analyze?
    - Analyzed epidemiological data (daily cases) from January 20 - March 16, 2020, and transport data (Metro travel) in South Korea  between 2017 - 2020. 
    - Studied the differences in epidemiological dynamics between Daegu (central to the SK outbreak) and Seoul. 
    - Inferred the time-dependent reproductive number (R_t) in each location. 

- What methods did they use?
    - Inferred time-dependent backward onset-to-confirmation delay distributions from daily case data using negative binomial regression. 
    - Used weakly informative priors for fixed effects ~ Normal(0,2)
    - Incubation period modeled by a Gamma dist, generation time by Negative Binomial

- Does this paper study COVID-19, SARS-CoV-2, or a related disease and/or virus?
    - Epidemiology of COVID-19

- What is the main finding (or a few main takeaways)?
    - social distancing and contact tracing in Daegu were effective (at least in part) at reducing the number of new cases 
    - Suggests that the COVID-19 epidemic can be suppressed with less extreme measures than those taken by China 
    - R_t in Daegu is now well below 1, however in Seoul is still above 1, this could lead to secondary outbreaks 
[...]

PR #120, opened 2020-03-30
Add Park2020_distancing manual reference

agitter:
I'm adding a manual reference for #109.  I need to test whether I set up the `.bib` file correctly before this can be merged.


agitter:
It worked!  From the artifact:
image
image

I did this by downloading the PDF from GitHub, importing the PDF into Zotero to extract the metadata, exporting to .bib, and adding the tag.

PR #131, opened 2020-04-01
Standardize headers / trailing whitespace

cgreene:
The implications of adjusting the trailing whitespace here might be profound on outstanding PRs and their merge process. We're going to have to bite the bullet on that at some point, but it'd be good to see what's outstanding and how painful that merge will become.


cgreene:
List of changes:
* Remove trailing whitespace (editor does this automagically, for better or worse)
* Remove pagebreak before sections - was annoying now when I was scrolling through the PDF
* Standardize header levels to match the Drug Name / Anticipated Mechanism / Current Evidence / Summary framing.
* Move a few things in the Avigan section because it didn't have all of these headers.

Issue #172, opened 2020-04-08
Making our Data More Accessible (for Internal and External Use)


rdvelazquez:
**Is your feature request related to a problem? Please describe.**
This review paper has multiple sources of information:
- The paper itself being the main source
- Paper summaries and reviews in issues
- Discussions in Pull-Requests  

This information isn't easily accessible in a programatic way.

It would be nice to include this information in the topic first paper tracking the @rando2 proposed in #144 but the information from our project isn't easily accessible

**Describe the solution you'd like**
Create a dataset of papers that we've looked at with information like:
- Link to GitHub issue where paper is summarized/reviewed
- If the paper is cited in our review paper (maybe with links to those sections or excerpts)
- Links to the Pull-Requests where the paper was involved

A TSV file with general paper columns (DOI, title, author, date) and project specific columns (issue links, pr links, paper links) would probably be fine. Maybe also include an excel file for easy download so those less technically inclined aren't scared away.

I've already started looking at getting some of this information and it does't look too hard:
- GitHub issues: https://github.com/rdvelazquez/covid19-review/blob/paperTracking/litsearch/python_Functions/Get%20Paper%20info%20from%20Github%20Issues.ipynb and https://github.com/rdvelazquez/covid19-review/blob/paperTracking/litsearch/issues.tsv
- Covid19-review citations: https://github.com/rdvelazquez/covid19-review/blob/paperTracking/litsearch/python_Functions/Get%20Info%20from%20Review%20Paper%20Citations.ipynb and https://github.com/rdvelazquez/covid19-review/blob/paperTracking/litsearch/citationsFromCovid19-Review.tsv

**Describe alternatives you've considered**
- We could just include this information in the topic specific TSV files proposed in #144 but this would be less useful for general use
- We could create a legit database but this seems like overkill for our project

**Additional context**
Relates to #96 #144 and #163

rdvelazquez:
@rando2 and anyone else: Let me know your thoughts on this. If we think it's something that would be useful I'll work on opening up a pull request with the code to generate a comprehensive TSV file on each Travis build. We could then look at how to syncs this with #144 #163 and any other sources of info.

rando2:
Hi @rdvelazquez, this is amazing because I spent all last night trying to figure out how to do this, but I use APIs approximately once every five years, so everything I know appears to be out of date and I did not make a ton of progress! 

[...]

PR #163, opened 2020-04-07
add initial mssm import

cgreene:
### Description of the proposed additions or changes

The folks at Mt. Sinai immunology have been doing a very nice job of reviewing the literature. These reviews are online with a compatible license, and I've been in touch with the folks there and they would be happy to have us include them as an appendix. We could refer to these and they would give more context to papers that we're already including here.

### Related issues

I have made a few PRs into https://github.com/ismms-himc/covid-19_sinai_reviews to standardize things to align with headers/citation styles that work with Manubot. I think we're at a reasonable place now. I've also supplied @cornhundred with a template for new reviews that should cleanly convert to manubot format.

### Suggested reviewers (optional)

Probably @agitter or @rando2 to see about how this is being integrated. I thought about plugging it into CI, but there's enough variability that I think the actual path will be more like:

1. Pull new versions of reviews.
2. Build locally and check.
3. File a PR with fixes into https://github.com/ismms-himc/covid-19_sinai_reviews
4. After @cornhundred merges those, pull new version & build locally to check.
5. Re-commit the appendix document and PR into our master here.

It'll also be good to hear from @cornhundred about how the collective author is being handled. We will also invite folks from their team to contribute individually to the review and/or to add themselves as individual authors.

I think we'll also want to get a table with:
* Author Names
* Author ORCIDs
* Author Conflicts of Interest
So that we have our ducks in a row for the collective author when time rolls around. We could format this as a markdown table if we wanted, but a CSV is probably easier for everyone (and manuscript submission systems also usually accept them).

### Checklist

- [ ] Text is formatted so that each sentence is on its own line.

- [ ] Pre-prints cited in this pull request have a GitHub issue opened so that they can be reviewed.

PR #144, opened 2020-04-04
Add Topic-First Tracking of New Papers

rando2:
### Description of the proposed additions or changes
The number of preprints is exploding, and we want to encourage communities forming around shared interests (e.g., #92 ) so that we can help move people from #17 onto specific components of the project.

The goal here is to help people find papers of interest and see if anyone else has already read that paper -- for example, a paper evaluating administration of Remdesivir + hydroxychloroquinine would be of interest to writers working on sections about both drugs.

In this PR, you'll find the code to automatically scan the [CDC's COVID-19 Research Article Database](https://www.cdc.gov/library/researchguides/2019novelcoronavirus/researcharticles.html) where they're already systematically identifying everything related to COVID-19 (thanks, @dziakj1 , this was a game-changing suggestion!) and select papers related to individual topics of interest.

### Demo Product
[Dummy Hub: Remdesivir](https://htmlpreview.github.io/?https://github.com/rando2/covid19-review/blob/AddTracker/litsearch/bytopic/Remdesivir.html)
[Dummy Hub: Tocilizumab](https://htmlpreview.github.io/?https://github.com/rando2/covid19-review/blob/AddTracker/litsearch/bytopic/Tocilizumab.html)

### Workflow
Run [`Run_Updates.R`](https://github.com/rando2/covid19-review/blob/AddTracker/litsearch/Run_Updates.R) to:
1. Look at the CDC website to find the most up-to-date list of COVID-19-related papers
2. Pull from [`keywords.txt`](https://github.com/rando2/covid19-review/blob/AddTracker/litsearch/keywords.txt) to filter for papers that use relevant words in the Title, Abstract, or Keywords
3. Write TSV file of relevant article for each therapy
4. Generate .Rmd files that display information in the TSV file corresponding to each therapy
5. Run the .Rmd file to use knitr to create a (user-friendly) HTML-based Topic Hub

### Additional Development Needed
1. We would need to determine the keywords we want to use. The ones I used are just an example.
2. I need to adjust the workflow so that we can update the TSV file through PRs to add issue numbers, etc. and then not lose that information when I check for updates. This would allow for collaborative updating of the tracking, but the HTML files won't automatically regenerate.
3. Originally I was picturing a Jupyter notebook so that anyone could run updates, but haven't quite figured out how to implement that without potentially introducing problems. 
[...]

PR #174, opened 2020-04-09
Create TSV with paper info

rdvelazquez:

A script to create a TSV file for each paper with link to GH issue, GH issue tags, link to citation in our paper and other information

### Related issues

#172 #144  #163

@rando2 @agitter 

### Checklist
- [ ] Get the Pull-Requests where papers were discussed (this might me more difficult than the issues and we may want to save this for later in a separate PR)
- [x] Merge the datasets (issues citations and PRs) _I'm just merging on DOI (not title) but this seems to work quite well_
- [x] Tie this in to CI ~or GH Actions~ (see: https://github.com/greenelab/covid19-review/pull/174#issuecomment-612431267)


litsearch/getInternalData.py
```diff
@@ -0,0 +1,160 @@
...
+        return citation
+
+    except:
+        print(
+            "the citation could not be automatically extracted from the following issue: \n",
+            issue["title"])
+        return "unknown"
+
+
+def getPaperTitleFromIssue(issue):
```
agitter, line 49: If we are able to extract the citation, we could again use Manubot to get the title.

litsearch/getInternalData.py
```diff
@@ -0,0 +1,160 @@
...
+            "https://api.github.com/repos/greenelab/covid19-review/issues?state=all&per_page=50&page=" +
+            str(pageNumber))
+        issues_page = json.loads(issuesResponse.text)
+        issues = issues + issues_page
+        numberOfIssuesReturned = len(issues_page)
+        pageNumber += 1
+    return issues
+
+
+def getCitationFromIssue(issue):
```
agitter, line 26: Most of our issues have DOIs.  If we wanted to be more general, we could use the Manubot package API to convert the citation (in any form: URL, PMID, DOI, etc.) to a CSL Item and then extract the DOI from the CSL if it exists.  That will be slower and not provide much benefit if the vast majority of issues have DOIs.

There are some corner cases like #109 that have atypical citations.

litsearch/getInternalData.py
rdvelazquez: This could be helpful. Would I call `manubot cite` via a subprocess or is there other python integration?

litsearch/getInternalData.py
rdvelazquez: This could be cool but I agree with you that it wouldn't provide much benefit at this point.

We can try to list the DOI in the issues and if this becomes problematic (i.e. the number of new paper issues being submitted by others increases and manually going into issues and putting the DOI becomes too much) we can revisit this.

rdvelazquez:
@rando2 This is starting to come together. A few questions for you:
1. Where should we save the TSV file (output branch or main branch? in the `litsearch` directory?)
2. Do we want to include this in CI on each build:
   - The downside of doing it with CI on each build vs using GH actions is that it won't reflect new issues (posted after the last build)
   - I still think this is preferable to trying to integrate into GH actions just for simplicity sake plus it seems like the builds are happening pretty regularly
3. Does the TSV file have all the info that you think we would want?

This could be improved (fuzzy match titles to resolve items that don't have DOIs, use Manubot for somethings, include PR info) but I think it will be useful as is. So if this is a blocker for #144 you could probably use it as is.

Once you have other sources that you would like to combine with this data you could use this function (https://github.com/greenelab/covid19-review/pull/174/files#diff-5f7a1c80e645be3cc036ba483e8db4b5R159) to do so, assuming the other sources have DOIs.

rando2:
@rdvelazquez so sorry I've been slow to get back to you! I'll take a look with fresh eyes in the morning and answer your questions.

litsearch/getInternalData.py
agitter: @rdvelazquez the manubot Python package has some importable functions you can use instead of calling the `manubot cite` command as a subprocess.

The best function is [`citekey_to_csl_item`](https://github.com/manubot/manubot/blob/a4c826ceb954635810ba204a2393dce6fa7a5d5a/manubot/cite/citekey.py#L278).  This takes in a Manubot-style citekey like `doi:10.1101/2020.04.08.20054551`.  It works for all valid citekeys (e.g. PMIDs, URLs) so if you ever needed to expand beyond DOIs you could use the same function.

The [test cases](https://github.com/manubot/manubot/blob/a4c826ceb954635810ba204a2393dce6fa7a5d5a/manubot/cite/tests/test_citekey_api.py) and (to a lesser extent) [CSL Item docs](https://manubot.github.io/manubot/reference/manubot/cite/csl_item/) will give you a sense of how to use the returned CSL Item object.  The [DOI tests](https://github.com/manubot/manubot/blob/a4c826ceb954635810ba204a2393dce6fa7a5d5a/manubot/cite/tests/test_doi.py) show more ways to interact with the CSL Item you'll get back.

If the citation information in the issue is well-formatted (not a safe assumption), something like this should work:
```python
citekey = getCitationFromIssue(issue)
# citekey is now a string like doi:10.1038/s41579-018-0118-9
csl_item = citekey_to_csl_item(citekey)
title = csl_item["title"]
# if ever needed at a later date for cross-referencing (not all CSL Items have valid PMID of course)
pmid = csl_item["PMID"]
```

rando2: This is great to know about!

rando2:
Hi @rdvelazquez, this is coming together so nicely! 
1. I don't have any strong preferences on where to save it, other than not in the root repository (since I think that is already a bit cluttered and confusing for newcomers!) If it's an intermediate file, perhaps `output` would make the most sense? I'm happy to remove `litsearch` if it doesn't end up being useful, I was just doing my development in a separate folder locally.
2. We build so often right now that I think that would be totally fine to just run CI on each build! Right now it's usually multiple times per day.
3. The TSV looks good to me! It's a little hard to predict what will happen once it's available to everyone, but I think DOI and title are the most critical elements. 

The other data source I would consider integrating is the list of reviews put together by @cornhundred and colleagues at Mount Sinai. They have a [google doc](https://docs.google.com/spreadsheets/d/e/2PACX-1vRngfhDKqZUEhHuQY60n3Bh76gkMQKeOq6D7UYkSgt0KPP7rcCTE-PjMeWO1g1YlGVhBTAMJS6rn-pc/pub?gid=0&single=true&output=tsv) but it seems like it might be easier to just use `GET /repos/:owner/:repo/contents/:path` to list all the files in this [directory](https://github.com/rando2/covid-19_sinai_reviews/tree/master/markdown_files) as they are named in a standard way according to DOI! Looking at your code, I feel like I might actually be able to help with adding in API functions-- 2015 strategy seems to hold up, even though my libraries needed some updating!

litsearch/getInternalData.py
rdvelazquez: Very cool; that's exactly what I was looking for. I'll give `citekey_to_csl_item` a try.

build/litsearch/getInternalData.py
```diff
@@ -0,0 +1,233 @@
...
+citationsData = getCitationsData()
+print(len(citationsData), "citations in the covid19-review paper")
+# TODO: pullRequestsData = getPullRequestData()
+
+print("\n -- merging the data --")
+combinedData = mergePaperDataFrames([citationsData, issuesData])
+print(len(combinedData), "total items after merge")
+
+
+combinedDataFilePath = "./output/sources_cross_reference.tsv"
```
agitter, line 231: Storing this in the `output` directory of the `output` branch makes sense to me.  Committing files back to the `master` branch gets messy.

If later we decide to include this as a new appendix, we could move it to the versioned website directories in the `gh-pages` branch.

rdvelazquez:
@agitter: [sources_cross_reference.tsv.zip](https://github.com/greenelab/covid19-review/files/4465630/sources_cross_reference.tsv.zip)

build/litsearch/getInternalData.py
```diff
@@ -0,0 +1,233 @@
...
+                merged[col] = merged[col].combine_first(df[secondaryCol])
+                merged.drop(secondaryCol, axis=1, inplace=True)
+
+    return merged
+
+
+# Main
+print("\n -- getting issues data --")
+issuesData = getIssuesData()
+print(len(issuesData), "'New Paper' issues")
```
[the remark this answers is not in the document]
rdvelazquez: Good call. I'll add that in.

rdvelazquez:
@agitter any way to silence Manubot from logging all the warnings and errors?
- Here's where it starts doing it in the logs: https://github.com/greenelab/covid19-review/pull/174/checks?check_run_id=579465981#step:6:158
- Here's the part of the code where it's coming from: https://github.com/greenelab/covid19-review/pull/174/files#diff-ff4bccd88512565b5a643b5d5adb30ffR60

It's probably fine but I have it in a `try/catch` block and I'm ok with errors there so it's just muddying up the logs.

build/litsearch/getInternalData.py
```diff
@@ -0,0 +1,239 @@
+import pandas as pd
+import requests
+import json
+import base64
+from manubot import cite
```
agitter, line 5:
```suggestion
import logging
from manubot import cite
```

build/litsearch/getInternalData.py
```diff
@@ -0,0 +1,239 @@
...
+            secondaryCol = col + "_" + str(i)
+            if secondaryCol in merged.columns:
+                merged[col] = merged[col].combine_first(df[secondaryCol])
+                merged.drop(secondaryCol, axis=1, inplace=True)
+
+    return merged
+
+
+# Main
+print("\n -- getting issues data --")
```
agitter, line 224:
```suggestion
# log only critical manubot errors
logger = logging.getLogger()
logger.setLevel(logging.CRITICAL)
print("\n -- getting issues data --")
```

cgreene:
@dasaderi pointed a resource out to me on twitter ( https://twitter.com/Neurosarda/status/1249363337114030080?s=20 ;  https://outbreaksci.prereview.org/ ) from PREreview. They have an API ( https://outbreaksci.prereview.org/api ), which might be the way to go if we wanted to link to or include information from their reviews. The URL structure might also be predictable (this is from an arXiv one): https://outbreaksci.prereview.org/2003.12444

Perhaps this should be a new issue - just raising it here first for consideration since this was the most relevant currently open effort :)

rdvelazquez:
@agitter - Thanks for the edits; the logs look much cleaner now.

@cgreene - That looks like another good source of reviews to try to sync up with our paper. We can look at that after we sync the Mt. Sinai reviews.

@rando2 - I'll let you merge this whenever is most helpful for you. I still want to try to get the pull requests related to each paper linked in the TSV but I may not get to that for a while so I can just open up a new PR for that if you want to merge this first.

rando2:
@rdvelazquez Excellent! I will update the branch and merge!

Issue #175, opened 2020-04-09
Maintaining the IIISM review appendix


agitter:
These ideas follow up on #163 to maintain the appendix of IIISM reviews:
- [x] Add update instructions to maintainers
- [x] Track the commit hash of the https://github.com/ismms-himc/covid-19_sinai_reviews repo to version the appendix (could be stored as a Manubot variable to appear in a link in the appendix)

PR #176, opened 2020-04-09
Minor fixes from #163

rando2:
Minor fixes for a whitespace issue and a header issue from #163


rando2:
The raw diff looks crazy, but the rich diff looks pretty normal, I'm not sure what's happening.
I'm going to re-do from command line.

cgreene:
I would not recommend editing this file. It is automatically generated and
will be overwritten.

rando2:
> I would not recommend editing this file. It is automatically generated and will be overwritten.

@cgreene I am happy to drop these small formatting issues! They really don't seem important! It's awesome that these two projects are connected now!

cgreene:
I am happy to make the changes upstream. Perhaps we should make a short
comment somewhere in the readme about how to make changes to this file?

agitter:
@cgreene that will be part of my instructions to maintainers in #175

rando2:
@agitter That would be great, apologies for misunderstanding this morning!

PR #190, opened 2020-04-11
Updates to contributions

rando2:
@cbrueffer @rdvelazquez @nilswellhausen @agitter  I noticed your roles were out of date because you've all been working on other areas! I batched these since this file is updated frequently.

@nilswellhausen I can't request review from you unless you comment so let's just say thumbs up/thumbs down if you think this looks ok. Everyone else, feel free to do that or approve.

### DIFF
--- a/content/metadata.yaml
+++ b/content/metadata.yaml
@@ -153,6 +153,7 @@ authors:
     contributions:
       - Methodology
       - Software
+      - Project Administration
     affiliations:
       - Department of Biostatistics and Medical Informatics, University of Wisconsin-Madison, Madison, Wisconsin, United States of America
       - Morgridge Institute for Research, Madison, Wisconsin, United States of America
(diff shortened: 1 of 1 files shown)


content/metadata.yaml
```diff
@@ -149,6 +152,7 @@ authors:
     email: gitter@biostat.wisc.edu
     contributions:
       - Methodology
+      - Software
```
rando2, line 155:
```suggestion
      - Software
      - Project Administration
```
@agitter I think this should be here too, I always struggle to map the CRediT taxonomy onto real life, but it feels like the people who are going through the PRs as they come in and coaching people through that process are all doing PA (which to date is primarily you, Christian, Nils and me)

nilswellhausen:
Thanks @rando2 👍

content/metadata.yaml
agitter: Thanks @rando2, that makes sense.  I agree that the CRediT taxonomy categories are subjective and hard to map precisely to different contributors.  You're doing a good job of monitoring contributions and applying them consistently to different contributors.

PR #204, opened 2020-04-14
Add PR Links to Internal Data TSV file

rdvelazquez:

Adds a link to any pull-requests related to each paper. Only looks at PRs with the "Text for Review" label. Includes a link if that paper's DOI was on a line of text that was touched (added, removed or edited) by the PR.

Here's what the TSV file looks like with this new addition: [sources_cross_reference.tsv.zip](https://github.com/greenelab/covid19-review/files/4477678/sources_cross_reference.tsv.zip)

### Related issues

#144, #172, #174 

@rando2 @agitter


rando2:
This looks amazing! Full disclosure, code review is a cornerstone of Casey's lab, but I started about a month and a half ago, so I'm really new to it and I'm not really sure what is the best way to give feedback.

I downloaded this branch and ran it locally. It did ran without error and did what it was supposed to do! The TSV opens so many possibilities!

The only thing I noticed that caught my eye was that the PR links use the following format:
<https://github.com/greenelab/covid19-review/compare/206>
If I go there, I actually don't see much. Changing "compare" to "pull" gives a URL that shows the changes associated with the PR:
<https://github.com/greenelab/covid19-review/pull/206>

I just thought I'd mention it in case this wasn't the intention!

rdvelazquez:
Thanks for looking at this @rando2! I think code reviews are great, both for keeping a code base in decent shape and for improving as a developer. So thanks for reviewing this and definitely provide any feed back you can. I give a big 👍 for doing legit reviews rather than the check the box type. 

> The only thing I noticed that caught my eye was that the PR links use the following format:
> https://github.com/greenelab/covid19-review/compare/206
> If I go there, I actually don't see much. Changing "compare" to "pull" gives a URL that shows the changes associated with the PR: https://github.com/greenelab/covid19-review/pull/206 which markdown treats as #206

> I just thought I'd mention it in case this wasn't the intention!

That was not my intention, good catch. On my end the links actually have the `pull` instead of the `compare`... I just ran it again to make sure that's what I was getting. Maybe there's something different about our environments? Rather than try to track down the difference it might be worth seeing what Travis CI gets; I don't know how to check that other than merging this in although there's likely a way to do that, not sure if it's worth it.

agitter:
This is another great feature.

> Rather than try to track down the difference it might be worth seeing what Travis CI gets; I don't know how to check that other than merging this in although there's likely a way to do that, not sure if it's worth it.

I was going to modify the GitHub Actions workflow to add save ` sources_cross_reference.tsv`, but then I realized it's already there!  For anyone following along, [these instructions](https://github.com/greenelab/covid19-review/blob/master/INSTRUCTIONS.md#how-can-i-see-my-change) show how to download the artifacts zip file that contains this tsv file.

I didn't review this pull request yet but thought this would be useful for checking environment-dependent behavior.

rdvelazquez:
Very good to know @agitter; thanks for pointing out how to see the build from a PR.

I checked the TSV file from CI artifacts and it looks like it has `pull` instead of `compare`:

Screen Shot 2020-04-15 at 8 44 54 AM

So we should be good on that front.

rdvelazquez:
@rando2 are you cool with only using PRs with the `Text for Review` tag? I didn't want it to include all PRs because some (reformatting, manual citations, etc.) aren't likely to have anything meaningful about the paper review. We just may want to go back and make sure the older PRs are appropriately tagged.

rando2:
Hi @rdvelazquez that sounds good to me! I try to be consistent about adding that tag and I think I get it on almost if not all of them! We just might want to loop in @ypar and @cbrueffer who are sometimes a step ahead of me!

rando2:
@rdvelazquez thanks for checking Travis CI! I think I just did something weird the first time because now I'm getting the right results!

rdvelazquez:
Ready by me.

cbrueffer:
Looks good!

build/litsearch/getInternalData.py
```diff
@@ -175,14 +175,51 @@ def dateStringFromDateParts(row):
...
+# Pull-Request Functions
+def getPRsFromAPI():
+    """ Gets all the pull-requests; Needs to use pagination because of 100 per-request api limit
+    """
+    PRs = []
+    pageNumber = 1
+    numberOfPRsReturned = 1
+    while numberOfPRsReturned != 0:
+        PRsResponse = requests.get(
+            "https://api.github.com/repos/greenelab/covid19-review/pulls?state=all&per_page=50&page=" +
```
agitter, line 187: Do you want to ignore pull requests that were closed without being merged?  Open and merged pull requests would be most relevant to track.

build/litsearch/getInternalData.py
```diff
@@ -221,6 +258,24 @@ def mergePaperDataFrames(dataFramesList):
...
+def addPRLinks(sourcesDF, prData):
+    """ Adds a new column to a sourcesDF with the link to any relevant PR that has that DOI """
+    def addPRLinkToPaperRow(row):
+        prLinks = []
+        doi = str(row.doi)
+        if len(doi) < 1:
+            return
+        else:
+            for PR in prData:
+                if doi in PR["diff"]:
```
agitter, line 270: If the text uses the short DOI, it could be missed here.

build/litsearch/getInternalData.py
rdvelazquez: I think it depends on what the data in the table ends up being used for. It might be good to include all the relevant PR's including those that weren't merged so that, for instance, if someone proposed changes but the changes were rejected that information and discussion would still be linked to the paper. With the goal of trying to be as open as possible I think keeping all PR's may be the better bet.

agitter: That logic makes sense to me

PR #207, opened 2020-04-14
Enable automatic section numbering

### LINKED ISSUE #205: Automatic Section Numbers
> Pandoc's manual states:
> image
> 
> I haven't tested this.  My reading is that automatic numbering is supported, in which case I definitely agree we should use it instead of manually updating these.  I'm not sure why only HTML is supported.  Hopefully because we're using athenapdf to convert the HTML manuscript to PDF, we would get the section numbers.

_Originally posted by @agitter in https://github.com/greenelab/covid19-review/diffs/4_

Thanks @agitter! #200 is not merged yet but I wanted to put this issue together in anticipation of it being merged soon. I will ping you when it goes through.

If you would be able to implement this, it would be hugely helpful. Some of the sections are very dense, and having more explicit labeling might help writers and readers. Also, @cgreene pointed out that this will make it easier to catch places where our headers are formatted incorrectly!

agitter:
Closes #205


agitter:
Modifying the Pandoc defaults file worked.  Pandoc numbers some sections that we would rather not, like the Authors.  I'm testing some other Pandoc settings, but you can check the artifacts now to see how it looks.

agitter:
I'm happy with the latest combination of numbering settings.  I found a way to disable numbering on some sections.

rdvelazquez:
Section numbering will be nice! Is there a way to make it start at `1` rather than `1.1`? 

Currently it's `1.1 Introduction`, `1.1.1 General Background` ..., `1.2 Pathogenesis` ...
But `1 Introduction`, `1.1 General Background` ..., `2 Pathogenesis` ... makes more sense to me.

agitter:
> Section numbering will be nice! Is there a way to make it start at 1 rather than 1.1?

There is a way to adjust that.  `manuscript-1.0.417-02b1c86.html` above shows what it would look like, though the actual numbering was broken in that build.  There are disagreements between how Manubot and Pandoc treat h1 and h2 HTML headers.  Manubot wants to use it for the title only so they are centered in the CSS.  Pandoc uses them as section headers.  https://github.com/manubot/rootstock/issues/273 discusses a related topic.

I believe that if we wanted to start the Introduction at 1 instead of 1.1 we would have to adjust the CSS.  I'm open to that but would want to save it for a later issue and pull request.  I'm not good with front-end tweaks.

rdvelazquez:
Gotcha; makes sense. And saving for later sounds good! 
I'll try to take a look at a workaround when (if) I have time.

Github won't let me review but this LGTM.

agitter:
If any of us do work on that later, it would likely involve:
- using the Pandoc `shift-heading-level-by` setting from 02b1c8658c8660cad2471ac27746fdfaa3338aeb
- modifying some of `default.html` so that h1 headers aren't centered, probably these lines https://github.com/greenelab/covid19-review/blob/495fac8a6d870ab62516d44c2cd04c64e7c25365/build/themes/default.html#L70-L96
- centering the h1 header with the title class:
```html
<h1 class="title">SARS-CoV-2 and COVID-19: An Evolving Review of Diagnostics and Therapeutics</h1>

rando2:
> Gotcha; makes sense. And saving for later sounds good!
> I'll try to take a look at a workaround when (if) I have time.
> 
> Github won't let me review but this LGTM.

@rdvelazquez sorry about that, I upped you to maintainer so hopefully you won't run into this issue again!

agitter:
@rdvelazquez this pull request was still open so I implemented your numbering suggestion.  How does `manuscript-1.0.477-9822c74.html` look?

This will also help with the manuscript's many levels of headers.  Each section header will be bumped up a size.

@rando2 does the latest build here look good to you aesthetically?

rando2:
> @rdvelazquez this pull request was still open so I implemented your numbering suggestion. How does `manuscript-1.0.477-9822c74.html` look?
> 
> This will also help with the manuscript's many levels of headers. Each section header will be bumped up a size.
> 
> @rando2 does the latest build here look good to you aesthetically?

@agitter Looks great! Thank you so much for doing this, I think it will really help to keep track of where we are in the broader structure. We have a lot of levels in therapeutics right but hopefully down the line we'll be able to streamline a bit. This will really help keep track of which sections go with what!

PR #215, opened 2020-04-15
Intro edit

ypar:

I edited the classifications section of the introduction. I clarified the virus description and added more information regarding how SARS-CoV-2 relates to other coronavirus strains. The remaining 'edits' are results of white-space removal.

Some notes:

* I have rearranged covid-19 portion of the general background to the classification section. The primary reason for rearranging this is the tone of writing. The general background starts with a laundry list of current events with dates and the covid-19 section is primarily scientific, especially with my edits. Although covid-19 description is not quite `classification`, it will be better described in relation to the descriptions regarding the virus.

* Some details I added can be omitted later. I think they are useful for now as we figure out how to structure the whole story.

* I deleted some `could be right but speculative` details. e.g. why mers and sars didn't become a pandemic could be due to containment, etc. but also could be due to differing biology (required concentration of infection, transmission rate, mode of transmission, etc.).

* I have mixed feelings about a list of specific dates, named geographic regions, etc. in a review article meant to focus on the disease and drugs, etc. Which `neighboring countries` should be listed, for example, is a bit subjective, which can be problematic.

* On a related note, a graph of cumulated cases will be more informative then, for example, the first-event-first-report-then-one-week-later type of a flow. Just a personal preference though. It will also depend on where this review will go - remain in manubot vs submit somewhere, etc. - and can be re-visited later. 

* If a graph will be used, it can be pulled from the public data table so it can be updated with CI. Same thing with the case and fatality count. I opened [an issue](https://github.com/greenelab/covid19-review/issues/218).

### Related issues

N/A

N/A

### Checklist

- [X] Text is formatted so that each sentence is on its own line.

### DIFF
--- a/content/02.introduction.md
+++ b/content/02.introduction.md
@@ -34,7 +34,7 @@ Most coronaviruses are considered zoonotic viruses with little to no transmissio
 
 
 A major group of coronaviruses include human coronaviruses (HCoVs) strains associated with multiple respiratory diseases of varying severity, ranging from common cold to severe pneumonia, with severe symptoms mostly observed in immunocompromised individuals [@doi:10.1086/377612].
-Among known human coronaviruses, four human coronaviruses (HCoV-229E, HCoV-NL63, HCoV-OC43 and HCoV-HKU1) are globally present in the human population and are estimated to contribute to approximately one-third of common cold cases annually [10.3390/diseases4030026].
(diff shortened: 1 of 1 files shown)


ypar:
One potential review/discussion point is that whether to make the consistent referral to coronavirus `coronavirus` or `CoV` in the mx. Accordingly, `human coronavirus` or `HCoV`?

ypar:
I am adding a couple of reviewer suggestions. Only one approval is required but some comments would be helpful.

PR #203, opened 2020-04-14
Added to Primary Transmission and Viral Entry

anskelly:
Description of the proposed additions or changes:
I read and compiled information from several recent structure/Cryo-EM papers detailing the interaction between SARS-CoV-2 S protein and ACE2. 

Related issues:
N/A

Checklist
 Text is formatted so that each sentence is on its own line. [YES]
 Pre-prints cited in this pull request have a GitHub issue opened so that they can be reviewed. [N/A]

PR #121, opened 2020-03-31
Add R0 and transmission section to Pathogenesis

alavendelm:
Drafted section on R0 and transmission dynamics. Few notes 
 - would be better to render as subscripts for R_0 and R_t, not sure how to do so in markdown? 
 - I have tried to keep this short and readable by only including essential data, e.g. not reporting confidence intervals, and not going into methods, can expand, but maybe these details better elsewhere? 
 - tagging @johnbarton as someone who may be able to help review/comment on/add to this section

PR #216, opened 2020-04-15
Manual reference for doi:10.1101/2020.01.27.20018952

agitter:
Fixes issues in #121 and #203 with `doi:10.1101/2020.01.27.20018952`.


cbrueffer:
The rendered version looks fine now :+1:

PR #217, opened 2020-04-15
Only cross-reference sources on GitHub Actions

agitter:
This attempts to fix the problem in https://github.com/greenelab/covid19-review/pull/203#issuecomment-613712525 by only running the Python script in GitHub Actions workflow, not AppVeyor.  I'll need to let it run in both services to see if it worked.


rdvelazquez:
I'm wondering if it actually failed because of exceeding the GitHub api limit (it's 60 per hour based on your IP address I believe). Maybe AppVeyor assigns us one IP and GH actions has rolling IP addresses or some direct integration with the GitHub api? 

I'm not sure why else it would fail on AppVeyor but run fine on GitHub Actions.

I should really make the script fail gracefully in all instances so that builds don't get stopped because of this (regardless of the reason).

Your changes look good though.

PR #230, opened 2020-04-18
Updates to Nutraceuticals

RLordan:
@rando2 @cgreene  Hi folks, just new info on vitamin C. I would appreciate your review of the content. Have a great weekend!

### Checklist

- [x] Text is formatted so that each sentence is on its own line.

- [ ] Pre-prints cited in this pull request have a GitHub issue opened so that they can be reviewed.

Issue #228, opened 2020-04-18
New Paper (Other): [The severe acute respiratory syndrome coronavirus 2 (SARS-CoV-2) envelope (E) protein harbors a conserved BH3-like motif]


marouenbg:
Title: The severe acute respiratory syndrome coronavirus 2 (SARS-CoV-2) envelope (E) protein harbors a conserved BH3-like motif.

# General Information
## Please paste a link to the paper or a citation here:

Link: https://www.biorxiv.org/content/10.1101/2020.04.09.033522v1

## What is the paper's [Manubot-style citation](https://github.com/greenelab/covid19-review/blob/master/USAGE.md#citations)?

Citation: [@doi:10.1101/2020.04.09.033522]

### Is this paper primarily relevant to Background or Pathogenesis? 

- [ ] Background
- [x] Pathogenesis
- [ ] Methods

## Please list some keywords (3-10) that help identify the relevance of this paper to COVID-19

* keyword 1 Protein E
* keyword 2 T cells
* keyword 3 BH3

## Which areas of expertise are particularly relevant to the paper?

- [x] virology
- [ ] epidemiology
- [ ] biostatistics
- [x] immunology
- [ ] pharmacology
- [ ] other:

rdvelazquez:
@marouenbg This paper was actually already summarized here: #195. I'm marking thisas a duplicate and closing this issue. I'll look into why that issue was missed in the script that looks for open issues.

Issue #165, opened 2020-04-07
New Paper (Other): Discovery of a novel coronavirus associated with the recent pneumonia outbreak in humans and its potential bat origin


ajlee21:
Title: Discovery of a novel coronavirus associated with the recent pneumonia outbreak in humans and its potential bat origin

# General Information
## Please paste a link to the paper or a citation here:

https://www.biorxiv.org/content/10.1101/2020.01.22.914952v1

## What is the paper's [Manubot-style citation](https://github.com/greenelab/covid19-review/blob/master/USAGE.md#citations)?

Citation: doi:10.1038/s41586-020-2012-7

### Is this paper primarily relevant to Background or Pathogenesis? 

- [x] Background
- [ ] Pathogenesis
- [ ] Methods

## Please list some keywords (3-10) that help identify the relevance of this paper to COVID-19

* keyword 1 (replace me, copy and paste more than three if needed)
* keyword 2 (replace me, copy and paste more than three if needed)
* keyword 3 (replace me, copy and paste more than three if needed)

## Which areas of expertise are particularly relevant to the paper?

- [x] virology
- [ ] epidemiology
- [ ] biostatistics
- [ ] immunology
- [ ] pharmacology
- [ ] other:

# Summary

### Suggested questions to answer about each paper:
- What did they analyze?
- What methods did they use?
- Does this paper study COVID-19, SARS-CoV-2, or a related disease and/or virus?
- What is the main finding (or a few main takeaways)?
- What does this paper tell us about the background and/or diagnostics/therapeutics for COVID-19 / SARS-CoV-2?
- Do you have any concerns about methodology or the interpretation of these results beyond this analysis?

### Any comments or notes?

Paper looks at the origin of SARS-CoV2
@johnbarton mentioned he might be interested in adding to this section

rdvelazquez:
This is a duplicate of #106

Issue #106, opened 2020-03-29
New Paper (Other): A pneumonia outbreak associated with a new coronavirus of probable bat origin


rando2:
Title: A pneumonia outbreak associated with a new coronavirus of probable bat origin

# General Information
## Please paste a link to the paper or a citation here:

Link: https://www.nature.com/articles/s41586-020-2012-7

## What is the paper's [Manubot-style citation](https://github.com/greenelab/covid19-review/blob/master/USAGE.md#citations)?

Citation: [@doi:10.1038/s41586-020-2012-7 ]

### Is this paper primarily relevant to Background or Pathogesis? 

- [x] Background
- [ ] Pathogenesis
- [ ] Methods

## Please list some keywords (3-10) that help identify the relevance of this paper to COVID-19

* SARS-CoV-2
* zoonotic disease
* bats
* phylogenetics
* genome
* protein domains

## Which areas of expertise are particularly relevant to the paper?

- [*] virology
- [ ] epidemiology
- [ ] biostatistics
- [ ] immunology
- [ ] pharmacology
- [*] other: omics

# Summary
If you would like to submit a summary of the paper, please copy and paste the following into a comment.

### Suggested questions to answer about each paper:
- What did they analyze?
- What methods did they use?
- Does this paper study COVID-19, SARS-CoV-2, or a related disease and/or virus?
- What is the main finding (or a few main takeaways)?
- What does this paper tell us about the background and/or diagnostics/therapeutics for COVID-19 / SARS-CoV-2?
- Do you have any concerns about methodology or the interpretation of these results beyond this analysis?

rando2:
@ajlee21 Here is the paper that never says "strain"

PR #240, opened 2020-04-23
Handle DOIs in Brackets; Throw Informative Error for Duplicate New Paper Issues

rdvelazquez:

- Correctly parses DOIs from issue citations in brackets; this was the reason that #107 wasn't correctly associated with the paper (see: https://github.com/greenelab/covid19-review/issues/228#issuecomment-615933570)
- Throws an informative error when duplicate issues exist (two issues for the same paper); There should be a build failure on this PR because  #106 and #165 address the same paper. I'll mark one as a duplicate after this PR builds; I just want to keep them both open for now so we can see how it will look in CI. _(Thee two issues didn't previously cause problems because #165 was missed due to the brackets)_ (see: https://github.com/greenelab/covid19-review/pull/230#issuecomment-615933976)

### Related issues

#172 

@agitter

### DIFF
--- a/build/litsearch/getInternalData.py
+++ b/build/litsearch/getInternalData.py
@@ -92,6 +92,7 @@ def getDOIFromCitation(citation):
             DOI = citation.split(".org/")[1]
         elif citationContainsDOI(citation):
             DOI = citation.split("doi:")[1]
+            DOI = DOI.replace("]", "")
         elif citation == "unknown":
             DOI = "unknown"
         else:
@@ -244,6 +245,20 @@ def addMtSinaiReviewLinks(df):
     return df
 
 
+def getDuplicates(array):
+    seen = {}
+    dupes = []
+
+    for x in array:
+        if x not in seen:
+            seen[x] = 1
+        else:
+            if seen[x] == 1:
+                dupes.append(x)
+            seen[x] += 1
+    return dupes
+
+
 # Data merging function
 def mergePaperDataFrames(dataFramesList):
     """ Combine a list of paper dataframes into one.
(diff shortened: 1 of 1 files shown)


rando2:
Hi @rdvelazquez this looks totally reasonable to me, though I will defer to Tony! I like that you are aware of the possibility that we could end up with more than two issues for the same DOI, which seems like something that could happen! Are you imagining that we might need to add that into the error message down the road?

rdvelazquez:
Thanks for looking at this @rando2!

> I like that you are aware of the possibility that we could end up with more than two issues for the same DOI, which seems like something that could happen! Are you imagining that we might need to add that into the error message down the road?

I tried to make the error message fairly generic so that it could handle that situation. If there's more than one issue per DOI it just throws an error and then lists all the DOIs that had more than one paper. I haven't tested an instance of three plus issues for a single paper but it should handle that fine. 

I'll wait for @agitter's review.

build/litsearch/getInternalData.py
```diff
@@ -267,6 +282,12 @@ def mergePaperDataFrames(dataFramesList):
         dataFramesList_DOI.append(df[validDOI])
         dataFramesList_NoDOI.append(df[invalidDOI])
 
+    # Check if there are issues with duplicate DOIs
+    for df in dataFramesList_DOI:
+        duplicateDOIs = getDuplicates(list(df.index))
```
agitter, line 287: The pandas `duplicated` function could be used to get duplicate rows. However, this is single-purpose code, so I don't think it matters whether we use that or a specialized function.

build/litsearch/getInternalData.py
```diff
@@ -267,6 +282,12 @@ def mergePaperDataFrames(dataFramesList):
...
+        if len(duplicateDOIs) > 0:
+            raise ValueError('The following paper(s) has/have duplicate issues:', duplicateDOIs)
```
agitter, line 289: I agree that we want the build to fail if that are duplicates.  That will be a stronger signal to maintainers to close the duplicate issues.

PR #206, opened 2020-04-14
Add a dPCR section to diagnostics.

cbrueffer:
### Description of the proposed additions or changes

Digital PCR is a promising technology to overcome the problem of false negative SARS-SoV-2 detection, albeit the equipment is not as ubiquitous as that for RT-PCR.

### Suggested reviewers (optional)

@rando2 @aadattoli 

### Checklist

- [x] Text is formatted so that each sentence is on its own line.

- [x] Pre-prints cited in this pull request have a GitHub issue opened so that they can be reviewed.

PR #253, opened 2020-04-26
Continue expanding / updating diagnostics section

davidmanheim:
I added a number of new sources for testing, and updated a few points. 
(Tests are no longer restricted to only those with known origin,  and CDC no longer says pregnant women are high risk.)
Added short section on pooled tests, and included information on stool sample testing.

Significant further expansion of section and sources.

### Checklist

- [ x] Text is formatted so that each sentence is on its own line.

### DIFF
--- a/content/10.diagnostics.md
+++ b/content/10.diagnostics.md
@@ -10,6 +10,8 @@ A range of diagnostic approaches from a methodological standpoint are being or c
 ### Molecular Tests
 
 Molecular tests are often carried out from nasopharyngeal cavity samples, including throat wash and saliva [@doi:10/ggp4qx].
+They have also been shown to work in one study using stool samples [@doi:10.1002/jmv.25742].
+This has been applied to broader community surveillance rather than for patient diagnostics, in studies that tested wastewater. [@doi:10.1016/j.scitotenv.2020.138764]
 
 #### RT-PCR
 
@@ -27,6 +29,13 @@ The assay uses the N gene to screen patients, while the orf1b region is used to 
 In this case the test was designed to detect sarbecoviruses, of which SARS-CoV-2 is one.
 Considering that no other sarbecoviruses are currently known to infect humans, a positive test indicates that the patient is infected with SARS-CoV-2.
 However, this test is not able to discriminate the genetics of viruses within the sarbecovirus clade.
+
+#### Pooled PCR Testing Approaches
+
+Due to limited supplies and the need for more tests, a number of labs have found ways to enable pooled or similar tests.
+The first such result was a group at the Technion, which found they could pool up to 32 samples [@doi:10.1101/2020.03.26.20039438], followed by larger pooling in work utlizing slightly different methods at the University of Minnesota [@doi:10.1101/2020.04.02.022186].
+These approaches are PCR-based, but allow for more rapid scaling and higher efficiency for testing.
+
 
 ##### CRISPR-based detection
 
@@ -51,15 +60,16 @@ For example, PCR-based test can be highly sensitive, but in high-throughput sett
 3. Type of specimen, as it is not clear which clinical samples are best to detect the virus [@doi:10.1038/d41587-020-00002-2].
 4. Expensive machinery, which might be present in major hospitals and/or diagnostic centers but is often not available to smaller facilities [@doi:10.1126/science.abb8400].
 5. Timing of the test, which might take up to 4 days to give results [@doi:10.1126/science.abb8400].
+6. The availability of supplies for testing, including swabs and testing media, has been limited [@doi:10.1128/JCM.00512-20].
 
 False positives are a concern with this technique, as the guide RNA can recognize other interspersed sequences on the patient’s genome, resulting in a loss of specificity.
-
 Similarly, in tests that use CRISPR, false positives can occur due to the specificity of the technique, as the guide RNA can recognise other interspersed sequences on the patient’s genome.
+False negatives are also a concern when sample collection (which can be difficult for nasopharyngeal cavity samples,) is done improperly.
 
 ### Serological Tests
 
 Despite diagnostic tests based on the detection of the genetic material can be quite sensitive, they cannot provide information about the extent of the disease over time.
-Most importantly, they would not work on a recovered patient who defeated the virus.
+Most importantly, they would not work on a patient who has fully recovered from the virus at the time of sample collection.
 In this context, immunity tests are significantly more informative.
 Additionally, they can help scientists to understand why the disease has a different course among patients, as well as what strategy might work to manage the spread of the infection.
 Furthermore, serological tests hold significant interest at present because they can provide information relevant to advancing economic recovery and allowing reopenings.
@@ -81,11 +91,18 @@ The authors are now working to get the assay into clinical use [@url:https://www
 
 Importantly, false-positives can occur due to the cross-reactivity with other antibodies according to the clinical condition of the patient [@url:https://www.fda.gov/media/136625/download].
 Therefore, this test should be used in combination with RNA detection tests [@url:https://www.fda.gov/media/136625/download].
+Serological tests also require waiting until a patient has had an imune system reaction sufficient to detect, which means that these tests are of limited use for diagnosis in the early stages of an infection.
+The limitation on timing makes serological tests vfar less useful for enabling test-and-trace strategies.
 
 ### Possible Alternatives to Current Practices for Identifying Active Cases
 
-- Are there other approaches that have worked for diagnosing other viruses at a rapid pace in large numbers of people?
-- What are some approaches people are currently testing for detecting live viruses, especially SARS-CoV-2?
+Clinical symptoms are similar enough to other types of pneumonia that purely symptomatic diagnosis is unlikely to be sufficient.
+However, related approaches for diagnosis of cases exist, though most of these are only likely to identify serious and symptomatic cases.
+
+X-ray diagnostics have high sensitivity but low specificity in some studies. [@doi:10.1148/radiol.2020200642]
+Others have shown that some doctors seems to be able to better distinguish, though sensitivity in this case was lower [@doi:10.1148/radiol.2020200823].
+However, preliminary machine-learning results have shown far higher sensitivity and specificity from analyzing chest X-rays than was possible with clinical examination [@doi:10.1007/s13246-020-00865-4].
+Even using machine learning, this class of diagnostics are limited to confirming clinically severe cases, however, and will not detect asymptomatic or presymptomatic infections, which is critical for stopping spread.
 
 ### Challenges to Diagnostic Approaches
 
@@ -100,9 +117,7 @@ Specifically, the individuals at risk include:
 - people with severe symptoms
 - people showing mild symptoms that have been in contact with a person who has developed the infection
 - people with underlying health conditions
-- pregnant women
 
-However, this method of testing administration does not detect a high proportion of infections.
-This includes individuals that are asymptomatic (potentially able to infect others) and individuals that are able to recover at home.
-Furthermore, individuals for whom the origin of the infection is uncertain are not tested.
+However, this method of testing administration does not detect a high proportion of infections, and does not allow for test-and-trace methods to be used.
+Individuals that are asymptomatic (potentially able to infect others) and individuals that are able to recover at home are therefore often unaware of their status.
 For instance, a recent study from the Imperial College estimates that in Italy the true number of infections is around 5.9 million against the 70,000 detected as of March 28th [@doi:10.25561/77731].


content/10.diagnostics.md
```diff
@@ -81,11 +91,18 @@ The authors are now working to get the assay into clinical use [@url:https://www
 
 Importantly, false-positives can occur due to the cross-reactivity with other antibodies according to the clinical condition of the patient [@url:https://www.fda.gov/media/136625/download].
 Therefore, this test should be used in combination with RNA detection tests [@url:https://www.fda.gov/media/136625/download].
+Serological tests also require waiting until a patient has had an imune system reaction sufficient to detect, which means that these tests are of limited use for diagnosis in the early stages of an infection.
```
matfax, line 94: "Due to the long incubation times and delayed immune responses of infected patients, serological tests are insensitive for diagnosis in the early stages of an infection."

content/10.diagnostics.md
```diff
@@ -81,11 +91,18 @@ The authors are now working to get the assay into clinical use [@url:https://www
...
+The limitation on timing makes serological tests vfar less useful for enabling test-and-trace strategies.
 
 ### Possible Alternatives to Current Practices for Identifying Active Cases
 
-- Are there other approaches that have worked for diagnosing other viruses at a rapid pace in large numbers of people?
-- What are some approaches people are currently testing for detecting live viruses, especially SARS-CoV-2?
+Clinical symptoms are similar enough to other types of pneumonia that purely symptomatic diagnosis is unlikely to be sufficient.
```
matfax, line 99: "so that" would clarify that "that" doesn't refer to the pneumonia but to the sentence.

content/10.diagnostics.md
```diff
@@ -81,11 +91,18 @@ The authors are now working to get the assay into clinical use [@url:https://www
...
+However, related approaches for diagnosis of cases exist, though most of these are only likely to identify serious and symptomatic cases.
```
matfax, line 100: What kind of "related approaches"?
"even though most of them are only sensitive to symptomatic and severe cases."

content/10.diagnostics.md
```diff
@@ -81,11 +91,18 @@ The authors are now working to get the assay into clinical use [@url:https://www
...
+
+X-ray diagnostics have high sensitivity but low specificity in some studies. [@doi:10.1148/radiol.2020200642]
+Others have shown that some doctors seems to be able to better distinguish, though sensitivity in this case was lower [@doi:10.1148/radiol.2020200823].
```
matfax, line 103: Others have shown that some doctors seem to be able to distinguish better, even though sensitivity, in this case, was lower

content/10.diagnostics.md
```diff
@@ -81,11 +91,18 @@ The authors are now working to get the assay into clinical use [@url:https://www
...
+However, preliminary machine-learning results have shown far higher sensitivity and specificity from analyzing chest X-rays than was possible with clinical examination [@doi:10.1007/s13246-020-00865-4].
+Even using machine learning, this class of diagnostics are limited to confirming clinically severe cases, however, and will not detect asymptomatic or presymptomatic infections, which is critical for stopping spread.
```
matfax, line 105: Regarding a- and presymptomatic infections: COVID-19 can manifest in the lung tissue and X-rays even though COVID-19 infected don't notice it. X-rays could benefit in this case. Machine-learning could increase the sensitivity of the X-ray scanning. A theoretical benefit is not disproven. I would write it this way:
"Yet, X-ray machine learning can only detect asymptomatic or presymptomatic infections that show lung manifestations. Therefore, it is unlikely to hinder the spread significantly."

content/10.diagnostics.md
```diff
@@ -100,9 +117,7 @@ Specifically, the individuals at risk include:
 - people with severe symptoms
 - people showing mild symptoms that have been in contact with a person who has developed the infection
 - people with underlying health conditions
-- pregnant women
 
-However, this method of testing administration does not detect a high proportion of infections.
-This includes individuals that are asymptomatic (potentially able to infect others) and individuals that are able to recover at home.
-Furthermore, individuals for whom the origin of the infection is uncertain are not tested.
+However, this method of testing administration does not detect a high proportion of infections, and does not allow for test-and-trace methods to be used.
+Individuals that are asymptomatic (potentially able to infect others) and individuals that are able to recover at home are therefore often unaware of their status.
```
matfax, line 122: "that": who
"(potentially able to infect others)": Either explain in a separate sentence or too self-explanatory to mention.

"that are able to recover at home": Is it really the attribute that they are able recover at home that makes them unaware of their status? I think it's more a psychological phenomenon to wait things out until they are too worse to neglect. Who isn't able to recover at home anyways?

davidmanheim:
I will try to address these all ASAP.

content/10.diagnostics.md
```diff
@@ -10,6 +10,8 @@ A range of diagnostic approaches from a methodological standpoint are being or c
 ### Molecular Tests
 
 Molecular tests are often carried out from nasopharyngeal cavity samples, including throat wash and saliva [@doi:10/ggp4qx].
+They have also been shown to work in one study using stool samples [@doi:10.1002/jmv.25742].
+This has been applied to broader community surveillance rather than for patient diagnostics, in studies that tested wastewater. [@doi:10.1016/j.scitotenv.2020.138764]
```
rando2, line 14:
```suggestion
This has been applied to broader community surveillance, rather than for patient diagnostics, in studies that tested wastewater [@doi:10.1016/j.scitotenv.2020.138764].
```

content/10.diagnostics.md
```diff
@@ -28,6 +30,13 @@ In this case the test was designed to detect sarbecoviruses, of which SARS-CoV-2
 Considering that no other sarbecoviruses are currently known to infect humans, a positive test indicates that the patient is infected with SARS-CoV-2.
 However, this test is not able to discriminate the genetics of viruses within the sarbecovirus clade.
 
+#### Pooled PCR Testing Approaches
+
+Due to limited supplies and the need for more tests, a number of labs have found ways to enable pooled or similar tests.
```
rando2, line 35:
```suggestion
Due to limited supplies and the need for more tests, a number of labs have found ways to enable pooled or otherwise parallelized tests.
```
I am also not sure of exactly what the right word would be! Does "parallelized" do it? I was also thinking "multiplexed"?

content/10.diagnostics.md
```diff
@@ -28,6 +30,13 @@ In this case the test was designed to detect sarbecoviruses, of which SARS-CoV-2
...
+The first such result was a group at the Technion, which found they could pool up to 32 samples [@doi:10.1101/2020.03.26.20039438], followed by larger pooling in work utlizing slightly different methods at the University of Minnesota [@doi:10.1101/2020.04.02.022186].
```
rando2, line 36:
```suggestion
The first such result came from a group at the Technion, which found they could pool up to 32 samples [@doi:10.1101/2020.03.26.20039438], which was then followed by larger-scale pooling in work that utilized slightly different methods at the University of Minnesota [@doi:10.1101/2020.04.02.022186].
```
Are these both RT-PCR with minor methodological changes, or is something more fundamental changing? Just in case it's helpful, FYI we have a section about dPCR at #206 that should be merged into the manuscript quite soon! Also, I'm wondering if larger pooling is "larger-scale pooling", or maybe "higher-throughput pooling"?

content/10.diagnostics.md
```diff
@@ -81,11 +91,18 @@ The authors are now working to get the assay into clinical use [@url:https://www
 
 Importantly, false-positives can occur due to the cross-reactivity with other antibodies according to the clinical condition of the patient [@url:https://www.fda.gov/media/136625/download].
 Therefore, this test should be used in combination with RNA detection tests [@url:https://www.fda.gov/media/136625/download].
+Serological tests also require waiting until a patient has had an imune system reaction sufficient to detect, which means that these tests are of limited use for diagnosis in the early stages of an infection.
+The limitation on timing makes serological tests vfar less useful for enabling test-and-trace strategies.
```
rando2, line 95:
```suggestion
The limitations due to timing make serological tests far less useful for enabling test-and-trace strategies.
```

content/10.diagnostics.md
rando2: Per @matfax comment, I actually don't mind the that there, but I would suggest the following change to the first part of the sentence:
```suggestion
Clinical symptoms are sufficiently similar to other types of pneumonia that purely symptomatic diagnosis is unlikely to be sufficient.
```

content/10.diagnostics.md
```diff
@@ -81,11 +91,18 @@ The authors are now working to get the assay into clinical use [@url:https://www
...
 
 ### Possible Alternatives to Current Practices for Identifying Active Cases
 
-- Are there other approaches that have worked for diagnosing other viruses at a rapid pace in large numbers of people?
-- What are some approaches people are currently testing for detecting live viruses, especially SARS-CoV-2?
+Clinical symptoms are similar enough to other types of pneumonia that purely symptomatic diagnosis is unlikely to be sufficient.
+However, related approaches for diagnosis of cases exist, though most of these are only likely to identify serious and symptomatic cases.
+
+X-ray diagnostics have high sensitivity but low specificity in some studies. [@doi:10.1148/radiol.2020200642]
```
rando2, line 102:
```suggestion
X-ray diagnostics have been reported to have high sensitivity but low specificity in some studies [@doi:10.1148/radiol.2020200642].
```

content/10.diagnostics.md
rando2: Maybe something like "Other studies have shown that specificity varies between radiologists, though the sensitivity reported here was lower"
(Assuming that's consistent with what the study says, I didn't check it!)

content/10.diagnostics.md
```diff
@@ -100,9 +117,7 @@ Specifically, the individuals at risk include:
 - people with severe symptoms
 - people showing mild symptoms that have been in contact with a person who has developed the infection
 - people with underlying health conditions
-- pregnant women
 
-However, this method of testing administration does not detect a high proportion of infections.
-This includes individuals that are asymptomatic (potentially able to infect others) and individuals that are able to recover at home.
-Furthermore, individuals for whom the origin of the infection is uncertain are not tested.
+However, this method of testing administration does not detect a high proportion of infections, and does not allow for test-and-trace methods to be used.
```
rando2, line 121:
```suggestion
However, this method of testing administration does not detect a high proportion of infections and does not allow for test-and-trace methods to be used.
```
Is test-and-trace currently defined anywhere in the text? If you don't mind checking and including an explanation if it isn't defined yet, that would be really helpful!

content/10.diagnostics.md
rando2: Perhaps "who are not advised to seek in-person medical attention" or something? Right now I believe the CDC recommendation is to stay at home and call your doctor rather than going in for an appointment until you get very severe. There is supposed to be drive through testing, but my understanding is that it is still not very accessible (all based on the USA, it's definitely better in some other places!)

matfax: This topic might be more complex than one sentence can explain. And it's also an interesting and useful topic to elaborate. Unidentified cases drive the spread, after all. It will most likely depend on the country, its policies, and the socio-cultural environment. Here are some aspects I spontaneously figured:
* Insufficient Capacities: What are the testing capacities? How well is the medical system equipped? How well can contact tracing be applied?
* Incomplete and inaccurate case identification: Is everyone allowed to be tested? Several countries only test cases with a traceable infection history. Some countries only test severe symptoms. Different testing methods are used.
* Financial insecurity: Is the test being paid for? Is follow-up health care being paid for? Does the government compensate for the missed wages?
* Improper sensitization: How do the government, local scientists, and the press sensitize the topic? A downplayed risk assessment can desensitize people to not take it serious enough. They play it safe, neglect it, might even spread it because they still think "it's just the flu" or because they start to believe some of the "conspiracy theories".

content/10.diagnostics.md
matfax: Sufficiently similar to be sufficient? I don't know...
Clinical symptoms are too similar to other types of pneumonia to be sufficient/reliable as a sole/single diagnostics criterion?

content/10.diagnostics.md
matfax: Multiplex PCR is something different, isn't it? How about "group testing"? What would be examples of these other pool-like tests?

content/10.diagnostics.md
davidmanheim: I will add that elsewhere in the next revision.

davidmanheim:
This is superceded with a different revision.

rando2:
@matfax and @davidmanheim you should also both make sure you [add yourself to the author list](https://github.com/greenelab/covid19-review/blob/master/CONTRIBUTING.md#authorship-guidelines)-- let me know if you have questions about how to do this!

content/10.diagnostics.md
rando2: Sorry I missed this before, @matfax! It is very interesting! I don't know how much public health we have/will have in the text, but these are all really interesting points.

PR #256, opened 2020-04-28
Become an author

davidmanheim:
### Checklist

- [x ] Text is formatted so that each sentence is on its own line.

- [x ] Pre-prints cited in this pull request have a GitHub issue opened so that they can be reviewed.


rando2:
We haven't been putting investigation for people who do the literature diving, but you're right that we probably should!

PR #261, opened 2020-04-29
Rootstock updates on 2020-04-28

agitter:
This pull request includes the latest Manubot Rootstock updates.  The biggest change is a spellchecker.  Funders are also now specified in a list format.

After this is merged, I'll make a follow up pull request to customize the dictionary for the spellchecker.


cgreene:
I think there might be something wrong with this. Is this saying there's a spelling error on line 3 in the abstract many times in a row?

```
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
content/01.abstract.md:3:CoV
```

agitter:
> I think there might be something wrong with this. Is this saying there's a spelling error on line 3 in the abstract many times in a row?

Yeah, that behavior is incorrect.  I couldn't reproduce it locally so I'm going to run a couple tests in the CI environment.

We shouldn't merge until this is resolved.

cgreene:
@agitter : I just got very rude and tried to see if I could figure out how to edit the dictionary. 😈

agitter:
2 new words in the dictionary eliminates 1000 spelling errors 👀

This seems to be working more reasonably now.  We might as well add a few more obvious words to the custom dictionary before merging so that it isn't so overwhelming when applied to new pull requests.

I'll take all current spelling errors, quickly review them, and add them to the custom dictionary if they aren't obvious mistakes.  If we notice typos that slipped through during manual editing, we can remove them from the dictionary.

cgreene:
@agitter : I'm not sure why the single characters (often spike protein, S) are still showing up. I see that you configured them to not show up. I just added them to the dictionary. We'll see if that helps...

cgreene:
Is it possible that configuration isn't getting used? I saw that you already had ignore case and lowercase `s` was already included as well. http://aspell.net/0.50-doc/man-html/4_Customizing.html#SECTION00522000000000000000

build/build.sh
```diff
@@ -82,6 +82,17 @@ if [ "${BUILD_DOCX:-}" = "true" ]; then
     --defaults="$PANDOC_DEFAULTS_DIR/docx.yaml"
 fi
 
+# Spellcheck
+if [ "${SPELLCHECK:-}" = "true" ]; then
+  export ASPELL_CONF="add-extra-dicts $(pwd)/build/assets/custom-dictionary.txt; ignore-case true; ignore 1"
```
cgreene, line 87:
```suggestion
  export ASPELL_CONF="add-extra-dicts $(pwd)/build/assets/custom-dictionary.txt;ignore-case true;ignore 1"
```

cgreene: maybe this is the issue and noting after the first parameter's actually getting used? 🤞

cgreene:
Still no luck 😠 . The environment variables must be working on some level, because the custom dictionary works. Do we know that the second environment variable is being used?

agitter:
The "S" errors are some corner case bug, or weird behavior at the very least.  The dictionary _should_ be case insensitive.  I tested the `ASPELL_CONF` ignoring words <= 5 characters in one of my manuscripts earlier, so I'm fairly sure that syntax is correct.

I'm running aspell directly on the original Markdown files to understand how the pandoc filter is using it and what is going wrong.  The first debugging runs hint that possessives are the problem with the "s".

cgreene:
At least one of the "re" cases was a hyphenation issue that I fixed over in #264. Some of the "S" ones might be possessive, but at least a few are the spike protein.

agitter:
The problem with the S is what is causing the pandoc spellcheck filter returns it as a spelling error in the first place when it is in the custom dictionary.  I'm very close to blaming it on a bug in the filter.

My test manuscript has the word "Pandoc's" in it and "s in the custom dictionary.  aspell on the original Markdown file returns "Pandoc's" as a misspelled word, as expected.  However, the pandoc spellcheck filter returns "s" as a misspelled word instead of "Pandoc's".

My attempt to read lua indicates the [pattern matching is broken](https://github.com/pandoc/lua-filters/blob/3c870cb5799fb1c4cb961b6648e5b3cddc50cfde/spellcheck/spellcheck.lua#L33). The pattern grabs the longest suffix that consists of all letters.  Thus the misspelled word "Pandoc's" is returned as "s".  Then, our post-processing uses grep to search for all instances of "s", which matches "(S)" and elsewhere.

If I'm right, this will require a workaround until fixed upstream.  We could remove all 1 and 2 character words from the misspelled words list, or re-process the misspelled words list by removing words that are in the dictionary.

agitter:
My fork of the pandoc spellcheck filter fixes the "s" problem.  It was a tricky error because the input Markdown can have `'`, but Pandoc converts it to `’` so the string patterns have to use `’`.

There still something similar happening with "re", but at least I understand how the spellcheck filter works now and how to fix it.

PR #265, opened 2020-04-30
Cache PR diff info to avoid exceeding GitHub API limit

rdvelazquez:

The `getInternalData.py` script makes calls to the GitHub API for:
- A list of all the issues
- A list of all the pull requests
- The files in the Mt. Sinai repo's markdown folder
- The "diff" of each "Text for Review" pull request  

This last item (the diffs) makes ~59 calls (one for each relevant PR) and that number will continue to increase. The GitHub api has a rateLimit of 60 calls per hour. The script was exceeding this limit and throwing an error (causing all Pull-Requests, see #206 for example, to fail).

This PR saves the results of the "diff" for closed PRs to a file so that it doesn't need to keep making those api calls. I tried to write it in a way that will work in a number of circumstances:
- Local vs. CI
- Without the cache file vs. with the cache file

### Related issues

#172 #257

@agitter


build/litsearch/getInternalData.py
```diff
@@ -200,19 +202,49 @@ def getRelevantPRData():
     diffHeader = headers.copy()
     diffHeader['Accept'] = "application/vnd.github.v3.diff"
     textForReviewPRs = []
+    
+    # Load the closed PRs info from a file (caching the closed PR info to avoid exceeding the GitHub API Limit)
+    closedPRFilePath = "./output/cache/closedPRDiffs.pkl"
+    if os.path.exists(closedPRFilePath):
```
agitter, line 208: Printing a message here will help us check whether it is working in CI.

rdvelazquez:
Thanks for reviewing this @agitter!

> You'll have to update the GitHub Actions workflow, either in this step or a new cache step

I'm actually proposing just using our own directory in the file system (in CI this will be the `cache` directory on the output branch; when running locally it will be the `output/cache` directory of whatever branch you are working in) to save the cache file rather than using the GitHub cache action. I'm recommending doing it this way because:

1. The GitHub cache action seems more designed for dependencies than other files (see: https://github.com/actions/cache/blob/master/examples.md#python---pip)
2. To avoid exceeding the api limit we need to take a "rolling" approach. i.e. only _updating_ the cache file not _regenerating_ it. This may be possible with GitHub's cache actions but it seemed more straight forward and transparent to do it with the file system.

Let me know if you think that's ok or if you think there's a reason and a way to use GitHub's cache action.

> We'll have to do some live testing to confirm it works as expected.

My expectation is that this pull request will build fine but it won't have the cache file available (we should be able to see that in the logs soon). Then after this is merged, subsequent builds will have the cache file available. That's what I see when I test it locally.

agitter:
Thanks for explaining your caching strategy in more detail.  I had assumed a different approach, but what you described should work.  The cache file would only be updated and stored when a new version of the manuscript is deployed, right?  That should be frequent enough.

I'll approve so that we can try it live.

PR #266, opened 2020-04-30
Meta: Author Acknowledgment

matfax:
As suggested by @rando2.

### DIFF
--- a/content/metadata.yaml
+++ b/content/metadata.yaml
@@ -442,3 +442,20 @@ authors:
     coi:
       string: "None"
       lastapproved: !!str 2020-04-28
+  -
+    github: matfax
+    name: Matthias Fax
+    initials: MF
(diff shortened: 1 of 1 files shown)


content/metadata.yaml
```diff
@@ -442,3 +442,20 @@ authors:
...
       string: "None"
       lastapproved: !!str 2020-04-28
+  -
+    github: matfax
+    name: Matthias Fax
+    initials: MF
+    orcid: 0000-0002-0064-395X
+    email: matthias.fax@gmail.com
+    contributions:
+      - Investigation
```
rando2, line 452:
```suggestion
```
For Investigation, I am assuming that will be used for the people who are doing the literature diving over in the issues. For now I think let's keep it for that purpose, but we can always add more roles to someone!

cgreene:
This is approved and build succeeded. Merging.

PR #274, opened 2020-05-01
Spellcheck improvements

agitter:
This update stores two spellcheck outputs as artifacts: the list of unique misspelled words and their locations in the markdown files.  The word list is intended to make it faster to update the custom dictionary.

It also uses my latest version of the lua figure that fixes bugs with accents and Unicode.  That should remove the "re" false positive spelling errors in the appendix here.


agitter:
is it okay to merge this?

PR #285, opened 2020-05-05
Auth api

rdvelazquez:

Trying to set the GitHub access token as an env variable for the script to use so the api limit will be 5,000/hr instead of 60/hr. I had tried caching the api responses and saving the results to a file(see #265) but this had issues because the the build script is run on the master branch but then the output is saved on the `output` branch... this is cleaner anyway.

### Related issues

#265 

@agitter


rdvelazquez:
@agitter I think this is working as expected on CI because:
- It tells us that the GitHub access token env variable is there: https://github.com/greenelab/covid19-review/pull/285/checks?check_run_id=646211770#step:6:133
- It doesn't exceed the api limit; i.e. this line is never logged: https://github.com/greenelab/covid19-review/pull/285/files#diff-ff4bccd88512565b5a643b5d5adb30ffR222

agitter:
The new use of the token looks good to me.

Do you want to keep the support for the cache file as well?  It complicates the function, but we may decide later we want it anyway to avoid hitting the GitHub API so frequently and redundantly.  To workaround the issue you found, we would have to do something like `git checkout output cache/closedPRDiffs.pkl` before running the script to retrieve the cache file from the output branch.

rdvelazquez:
> Do you want to keep the support for the cache file as well?

I'd prefer to keep it out for now for the sake of simplicity. We have the commit history so that code won't be lost if we decide to use it again in the future.

Issue #235, opened 2020-04-20
Topic-first identification of preprints


rando2:
@danich1, who is a PhD student in Casey's lab and an NLP expert, has developed an approach to categorize bioRxiv records using word2vec. 

You can find his project [here](https://github.com/greenelab/annorxiver/) and the notebook that shows how to (https://github.com/greenelab/annorxiver/blob/master/biorxiv/word_vector_experiment/biorxiv_generate_embeddings.ipynb) 

Note that (I believe) his approach also generates the author type (review vs research), heading (whether it's new results, contradictory results, etc.), and category (field) labels by examining the bioRxiv records. This seems like a way more elegant solution to what I was trying to do in #144, and David has very generously offered to help us adapt his approach for this project!

I wanted to loop in @rdvelazquez and @agitter who have been leading the preprint tracking. I had a few questions in [an issue](https://github.com/greenelab/annorxiver/issues/7) but assuming we can generate a similar dataset to what David is working from, I think it David's workflow and Ryan's workflow will integrate quite nicely!

rando2:
I'm putting this announcement of bioRxiv's XML repository here in case it's useful later: <http://connect.biorxiv.org/news/2020/04/18/tdm>

danich1:
Oh awesome find @rando2. Look like bioRxiv got their xml dump ready for the public. As answered in [greenelab/annorxiver#7](https://github.com/greenelab/annorxiver/issues/7) I couldn't share the dump as the bioRxiv group didn't want me to share it with anybody until they were ready to go public. 

BioRxiv is an option to go down, but it might be worth taking a look at [this dataset](https://allenai.org/data/cord-19). AllenAi constructed their own specific COVID-19 dump in json format. They also are hosting a [kaggle competition](https://www.kaggle.com/allen-institute-for-ai/CORD-19-research-challenge) using that dataset (kernels have cool results and have code that does embeddings as previously discussed). Avenues to consider, but if my code is ended up being used I can adapt it to AllenAi's dataset. Will be a simple fix.

rando2:
Thank you so much for this @danich1! So I looked into the structure of that Allen AI dataset you suggested and it seems pretty straightforward to combine the metadata file fwith the output from what Ryan has been building and use it in your approach. 

[...]

PR #226, opened 2020-04-17
Add Mt. Sinai review links to TSV file

rdvelazquez:
### Description of the proposed additions or changes

Adds a new column to the TSV file with link to Mt. Sinai review if available.

### Related issues

#212  #163 

### Suggested reviewers (optional)

@agitter @cgreene @rando2

PR #286, opened 2020-05-07
Combine Allen AI dataset with our referenced papers

rando2:

This code retrieves the Allen AI dataset, combines that data with the data tracked in output/sources_cross_references.tsv, and then pulls metadata from bioRxiv or medRxiv for any preprints that aren't in the Allen AI dataset yet. This includes the abstract, but not the full text.

The full bioRxiv XML dataset apparently only updates once a month, whereas Allen AI updates once a week, so unfortunately it probably isn't worth it to use the bioRxiv AWS interface to pull the full text. @danich1 I don't know if this plan works OK on your end. 

### Related issues

#235 
#226 

@rdvelazquez @agitter 

Note: I tried to follow the same relative path that @rdvelazquez's used, referring to the output directory as ./output. I assume that the CI runs out of the root covid19-review directory but just want to check that is right.

### DIFF
--- a/build/litsearch/combineDataSets.py
+++ b/build/litsearch/combineDataSets.py
@@ -1,0 +1,96 @@
+import pandas as pd
+import os
+import urllib.request
+import requests
+import datetime as dt
+
+def update_AI(fname):
+    # Download most up-to-date metadata from Allen AI COVID-19 dataset
+    url = 'https://ai2-semanticscholar-cord-19.s3-us-west-2.amazonaws.com/latest/metadata.csv'
+    urllib.request.urlretrieve(url, filename=fname)
+
+def clean_df(df):
+    df = df.dropna(subset=["doi"])
+    return df
+
+def paper_info(row):
+    """Queries metadata from bioRxiv or medRxiv based on paper DOI
+    Accepts: Row formatted as review_sources df, must have "doi" and "publication"
+    Returns: Row, updated as possible from *arxiv API """
+    if row["publication"] in ["bioRxiv", "medRxiv"]:
+        # Format URL for *rxiv API and convert metadata to dict from json response
+        pub = row["publication"].lower()
+        doi = row["doi"]
+        url = "https://api.biorxiv.org/details/{0}/{1}".format(pub,doi)
+        r = requests.get(url)
+        paper_data = r.json()
+        metadata = dict(paper_data["collection"][0])
+
+        # Update the row with metadata retrieved
+        row["URL"] = url
+        for kwd in ["title", "authors", "date", "abstract"]:
+            row[kwd] = metadata[kwd]
+    return row
+
+if __name__ == "__main__":
+    # Update external data, if not updated yet today, and read
+    metadata_fname = "../../output/AllenAI-metadata.csv"
+    if not os.path.exists(metadata_fname) or \
+            dt.datetime.fromtimestamp(os.path.getctime(metadata_fname)).date() != \
+            dt.datetime.now().date():
+        print("Update Allen AI metadata")
+        update_AI(metadata_fname)
+    AIA_metadata = pd.read_csv(metadata_fname, low_memory=False)
+    AIA_metadata = clean_df(AIA_metadata)
+    print("AIA has {0} sources".format(len(AIA_metadata)))
+
+    # Update internal dataset, which is maintained with CI
+    review_sources = pd.read_csv("../../output/sources_cross_reference.tsv",
+                                 sep="\t")
+    review_sources = clean_df(review_sources)
+    review_sources.rename({"URL":"SecondaryURL"}, inplace=True, axis=1)
+    print("COVID-19 review has {0} sources".format(len(review_sources)))
+
+    # Add tracking from project to the Allen AI dataset
+    combined = AIA_metadata.merge(review_sources,
+                                  on="doi",
+                                  how="outer")
+    print("Combined set has {0} sources".format(len(combined)))
+
+    # Remove duplicate columns
+    to_rename = dict()
+    for dup_col in ["title"]: # loop in case we end up with other duplicated info to remove
+        combined[dup_col + "_x"].update(combined[dup_col + "_y"])
+        combined.drop(dup_col + "_y", axis=1, inplace=True)
+        to_rename[dup_col + "_x"] = dup_col
+    combined.rename(to_rename, inplace=True, axis=1)
+
+    # Retrieve data for internally referenced dois not in Allen AI dataset
+    missing_dois = pd.Series(list(set(review_sources["doi"]).
+                                  difference(set(AIA_metadata["doi"]))),
+                             name="doi")
+    missing = pd.merge(combined,
+                       missing_dois.to_frame(),
+                       on='doi',
+                       how='inner',
+                       left_index=True)
+    missing.apply(paper_info, axis=1)
+    for source in ["bioRxiv", "medRxiv"]:
+        if len(missing[missing["publication"] == source]) > 0:
+            print("Abstract retrieved from {0} for these DOIs:".format(source))
+            print(missing[missing["publication"] == source]["doi"])
+
+    # Update combined with information from bioRxiv/medRxiv
+    combined.set_index("doi", inplace=True)
+    combined.update(missing.set_index("doi"))
+    combined['doi'] = combined.index
+
+    # Tabulate papers not in Allen AI and not available through bio/medRxiv API
+    num_pubs_missing = missing['publication'].value_counts().to_dict()
+    print("{0} unmatched references, {1} not in Allen AI, bioRxiv, medRxiv".format(
+        sum(num_pubs_missing.values()),
+        sum([num_pubs_missing[pub]
+             for pub in num_pubs_missing.keys()
+             if pub not in ["bioRxiv", "medRxiv"]])))
+
+    combined.to_csv("../../output/covid19-lit-all.tsv", sep="\t", index=False)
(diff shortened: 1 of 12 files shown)


build/litsearch/combineDataSets.py
```diff
@@ -0,0 +1,96 @@
...
+import requests
+import datetime as dt
+
+def update_AI(fname):
+    # Download most up-to-date metadata from Allen AI COVID-19 dataset
+    url = 'https://ai2-semanticscholar-cord-19.s3-us-west-2.amazonaws.com/latest/metadata.csv'
+    urllib.request.urlretrieve(url, filename=fname)
+
+def clean_df(df):
+    df = df.dropna(subset=["doi"])
```
rdvelazquez, line 13: Depending on how the resulting `covid19-lit-all.tsv` file ends up being used, we will want to keep in mind that sources without DOIs have been removed and may want to consider adding them back in.

build/litsearch/combineDataSets.py
```diff
@@ -0,0 +1,96 @@
...
+    AIA_metadata = pd.read_csv(metadata_fname, low_memory=False)
+    AIA_metadata = clean_df(AIA_metadata)
+    print("AIA has {0} sources".format(len(AIA_metadata)))
+
+    # Update internal dataset, which is maintained with CI
+    review_sources = pd.read_csv("./output/sources_cross_reference.tsv",
+                                 sep="\t")
+    review_sources = clean_df(review_sources)
+    review_sources.rename({"URL":"SecondaryURL"}, inplace=True, axis=1)
+    print("COVID-19 review has {0} sources".format(len(review_sources)))
```
rdvelazquez, line 52: FYI - I think this number will also include:
- papers from `New Paper` issues that aren't actually cited in COVID19-review
- papers that were reviewed by Mt. Sinai but not otherwise cited in COVID19-review

agitter:
Thanks for the new features @rando2 and the review @rdvelazquez.  I haven't reviewed in full, but I saw some of this involves reading files from the `output` subdirectory.  If we want that to work on CI, I can help strategize on how to modify the GitHub Actions workflow.

danich1:
@rando2 my pipeline will still work without the biorxiv aws. Just let me know when you are ready to use it.

build/litsearch/combineDataSets.py
rando2: @rdvelazquez that is a very good point! For @danich1 analysis I think we need to drop them, which means they wouldn't be among the sources suggested for each topic (but it would be hard to suggest them without being able to pull text). For anything involving internal curation, we should definitely stick with the original output from your script.

build/litsearch/combineDataSets.py
rando2, line 52:
```suggestion
    print("COVID-19 review has examined {0} sources".format(len(review_sources)))
```
Very true! Hopefully this disambiguates a bit? Also open to other suggestions about how to describe it!

build/litsearch/combineDataSets.py
```diff
@@ -0,0 +1,96 @@
...
+import urllib.request
+import requests
+import datetime as dt
+
+def update_AI(fname):
+    # Download most up-to-date metadata from Allen AI COVID-19 dataset
+    url = 'https://ai2-semanticscholar-cord-19.s3-us-west-2.amazonaws.com/latest/metadata.csv'
+    urllib.request.urlretrieve(url, filename=fname)
+
+def clean_df(df):
```
rando2, line 12:
```suggestion
```

build/litsearch/combineDataSets.py
rando2, line 13:
```suggestion
```

build/litsearch/combineDataSets.py
```diff
@@ -0,0 +1,96 @@
...
+    df = df.dropna(subset=["doi"])
+    return df
```
rando2, line 14:
```suggestion
```

build/litsearch/combineDataSets.py
```diff
@@ -0,0 +1,96 @@
...
+if __name__ == "__main__":
+    # Update external data, if not updated yet today, and read
+    metadata_fname = "./output/AllenAI-metadata.csv"
+    if not os.path.exists(metadata_fname) or \
+            dt.datetime.fromtimestamp(os.path.getctime(metadata_fname)).date() != \
+            dt.datetime.now().date():
+        print("Update Allen AI metadata")
+        update_AI(metadata_fname)
+    AIA_metadata = pd.read_csv(metadata_fname, low_memory=False)
+    AIA_metadata = clean_df(AIA_metadata)
```
rando2, line 44:
```suggestion
    AIA_metadata = AIA_metadata.dropna(subset=["doi"])
```

build/litsearch/combineDataSets.py
```diff
@@ -0,0 +1,96 @@
...
+    print("AIA has {0} sources".format(len(AIA_metadata)))
+
+    # Update internal dataset, which is maintained with CI
+    review_sources = pd.read_csv("./output/sources_cross_reference.tsv",
+                                 sep="\t")
+    review_sources = clean_df(review_sources)
```
rando2, line 50:
```suggestion
    review_sources = review_sources.dropna(subset=["doi"])
```

rando2:
@agitter and @rdvelazquez, any advice about how to integrate this into CI would be extremely helpful. I haven't contributed to the CI before, so I would be very excited to learn how to do this!

rdvelazquez:
> @agitter and @rdvelazquez, any advice about how to integrate this into CI would be extremely helpful. I haven't contributed to the CI before, so I would be very excited to learn how to do this!

I tried to trace what the CI actually does by following the steps here: https://github.com/greenelab/covid19-review/tree/master/.github/workflows which mostly just runs scripts from here: https://github.com/greenelab/covid19-review/tree/master/build

The fact that we want the resulting files to be saved to the output branch is a complicating factor that I think @agitter had some ideas about.

If you can tie this script in to the build.sh script that may be all that's needed for getting it in CI.

agitter:
As @rdvelazquez said, to run this script in the CI you will most likely want to call it in the `build.sh` script after this line https://github.com/greenelab/covid19-review/blob/8f84fb97ba83b794723757feb5474b010fb28205/build/build.sh#L106

That `build.sh` script runs in the "Build Manuscript" step of the GitHub Actions workflow
https://github.com/greenelab/covid19-review/blob/8f84fb97ba83b794723757feb5474b010fb28205/.github/workflows/manubot.yaml#L46

A separate script `ci/deploy.sh` runs during the "Deploy Manuscript" step of the workflow, but only on the master branch.  So for pull requests, it will only run after the pull request is merged, not on the CI builds that execute on each pull request commit.  Part of what the deploy step does is push files in the `output` subdirectory to the `output` branch of the repo.  For example, that's what happens with `sources_cross_reference.tsv` already.  So storing a file to the `output` branch is easy.

Retrieving a file from the `output` branch for caching purposes is more complicated and untested.  Some version of the `git checkout` command I suggested at https://github.com/greenelab/covid19-review/pull/285#issuecomment-625012541 may work, but we'll need to do some live testing to see what happens.  The goal would be to copy the cache file from the `output` branch to the current branch being tested in CI so that it can be loaded by your Python script.

rdvelazquez:
We might be able to avoid caching altogether. I incorporated your script into CI and just removed the caching. It now downloads the Allen AI meta data set on each run which isn't ideal but I think it's workable... the download is free for us, pennies for Allen AI, very fast, and (more critically) I think we should be fine as far as api limits from them. I looked at getting caching to work with CI when I was dealing with the GitHub api limit and it seemed like a bit of a pain (either using GitHub Action's caching or getting files from the output branch).

Here's the changes I did to get it into CI: https://github.com/rdvelazquez/covid19-review/commit/b4fee13342c1792f1a6b099675637cb7cdfa2fb2.

agitter:
I looked at the GitHub Actions caching again, and I should be able to set it up easily.  It's simpler than checking out files from the `output` branch.  How about we merge @rdvelazquez's changes from https://github.com/rdvelazquez/covid19-review/commit/b4fee13342c1792f1a6b099675637cb7cdfa2fb2 into this branch except for the cache removal.  Then I can try setting up the cache.

We won't need to cache `sources_cross_reference.tsv` because it is being written each time in the CI build.  I would cache `AllenAI-metadata.csv`.  Do we also want that file available in the `output` branch after deploying a new version of the manuscript, or is `covid19-lit-all.tsv` the only output file to track?

rando2:
Thank you both for looking at this! @agitter I don't think we need to track the Allen AI dataset in output. @rdvelazquez I added your branch to this PR but I didn't change the caching at all. I'm looking over #285 but am not confident about how much/which parts should be added back in. Would you be able to advise on this? I'm definitely learning a lot about CI and GitHub workflows from working with you both, thank you so much for your patience!

rdvelazquez:
> @rdvelazquez I added your branch to this PR but I didn't change the caching at all. I'm looking over #285 but am not confident about how much/which parts should be added back in. Would you be able to advise on this?

I think @agitter was referring to adding back in the caching of the Allen AI dataset that I removed here: https://github.com/greenelab/covid19-review/pull/286/commits/31993e95d9408da5b9ab44115a2b149ccdd5d200#diff-c5f6131f5a3bf820209d550ac1ccd30eL34

My recommendation would be to merge this in as is (without adding the caching back in) and then @agitter can work on adding caching (if that's something that you guys want) but it won't slow the progress of the other litsearch tasks that may be waiting on this PR. But obviously I'll leave that up to you.

rando2:
Wonderful, thank you @rdvelazquez! I'll merge this in, and then @danich1 hopefully we are ready for you whenever you get a chance! Please just let me know if any adjustments are needed!

agitter:
I agree this was okay to merge.  Caching won't impact functionality.

PR #311, opened 2020-05-13
addedReference.diagnostics.md

yemarshall:
I added for the simple formula a book reference I previously mentioned as a comment while adding the excerpt on herd immunity. I'm still learning manubot & my pull request for the herd immunity clarification may have been circuitous so i'm adding a new pull request for this reference rather than trying to open the previous pull request.

### Checklist

N/A


yemarshall:
Note on agitter note above. I think the way the citation looks as generated ~automatically is ok. Looking at my notes on the book, it seems that each chapter had its own authors so that May & MacLean would be the editors.

agitter:
Thanks for pointing that out @yemarshall.  That's great that the automated metadata retrieval for this book is correct.  The WorldCat [link](https://worldcat.org/isbn/9780199209989) I checked has the misled me regarding the author/editor info.

Issue #313, opened 2020-05-14
Questions on Reproduction Number Section


rdvelazquez:
@alavendelm and @johnbarton the [reproduction number section](https://greenelab.github.io/covid19-review/#reproduction-number-and-dynamics-of-transmission) that you wrote was really informative. I just had a couple questions after reading it and thought I'd share in-case they can help improve the document or if others had similar questions:

1. - [x] Do we think it’s worth noting the efforts at “live”predictions of Rt? I know of https://rt.live and https://epiforecasts.io/covid/ but there may be others.
2. - [x] "R0 and Rt can be estimated directly from epidemiological data or inferred using mathematical modeling: the susceptible-infected-recovered (SIR) model and its extensions [39](https://greenelab.github.io/covid19-review/#ref-lwH4gcts).” Is this the right citation for SIR model and extensions? It seemed like that paper was related to just the intra-host response.
3. - [x] Would a tabular or graphical summary be a useful addition? Maybe something like a row for each paper and columns for: location, date range the estimate covers, method used, point estimate, CI range and source/link. Something similar was done here: https://academic.oup.com/jtm/article/27/2/taaa021/5735319 but it might be a good addition in this review paper because it's a living document and could be updated as new data comes out. I could help out with preparing a table or graph summary if that's something that we want. **Edit: This was discussed below and could be a future addition but likely not in the near term**
4. - [x] The difference between R0 and Rt as used here wasn’t very clear to me. I think we are just using the terms that the cited papers used but the distinction seems arbitrary and might cause confusion. The second paragraph lists estimates of R0 but some of the estimates are explicitly for a specific populations at a specific time (which would make them Rt I believe). For example "on a cruise ship where an outbreak occurred, predicted R0=2.28 [46]” It seems like this should be an effective reproduction number (Rt) rather than basic reproduction number (R0).

rando2:
@rdvelazquez for point 3, @ypar has informed me that manubot can actually pull data and automatically update figures! It is on my to-do list to look into for replacing the beginning of the intro where we talk about case/death numbers, but I haven't gotten there yet. If it's something you're interested in, I can move it up in terms of priority!

alavendelm:
Hi @rdvelazquez, thanks these are v helpful comments. Point 3 (incorporating point 1) would be a great addition! We just need to think carefully about what data sources to use for it to make sure we trust their reliability enough. I will look into rt.live and the other source.

[...]

PR #352, opened 2020-05-29
clarification of variation in R0 estimates

rdvelazquez:

Adds some clarification on how to interpret the variability in estimated values of R0. @alavendelm as you wrote the initial draft... let me know if you think this is (1) correct and (2) a worthwhile clarification. If you think the variability in estimated R0 came more from the methods used than from the different populations then we could discard this suggested clarification or modify it in some way.

### Related issues

#313 

@alavendelm @yemarshall 

### Checklist


content/07.pathogenesis.md
```diff
@@ -55,7 +55,8 @@ When R~t~ is greater than 1, an epidemic grows (i.e., the proportion of the popu
 R~0~ and R~t~ can be estimated directly from epidemiological data or inferred using mathematical modeling.
 Modeling approaches are typically based upon a classic epidemiological model structure: the susceptible-infected-recovered (SIR) model and its extensions [@doi:10.1126/science.272.5258.74].
 
-R~0~ for COVID-19 is estimated to lie in the range R~0~=1.4-6.5 [@doi:10/ggjvr7; @doi:10.1093/jtm/taaa021; @doi:10.1126/science.abb3221]; estimates vary considerably depending on the data and the methods used.
+Estimates of R~0~ for COVID-19 lie in the range R~0~=1.4-6.5 [@doi:10/ggjvr7; @doi:10.1093/jtm/taaa021; @doi:10.1126/science.abb3221].
+Variation in R~0~ is expected between different populations and the estimated values of R~0~ discussed below are for specific populations at specific times, therefore the different estimates of R~0~ should not necessarily be interpreted as a range of estimates of the same underlying parameter.
```
rando2, line 59:
```suggestion
Variation in R~0~ is expected between different populations and the estimated values of R~0~ discussed below are for specific populations at specific times; therefore, the different estimates of R~0~ should not be interpreted as a range of estimates of the same underlying parameter.
```

rando2:
I am happy to merge but thought it might be good to ask @alavendelm to approve first!

alavendelm:
oh!! I thought I commented on this few days ago but I started a review and was left "pending"
now deleted 🤦 

I have a concern about the last clause - don't think it is necessarily true, there is, in theory, one R0 for coronavirus in a human population. Will suggest alternative

content/07.pathogenesis.md
alavendelm:
```suggestion
Variation in R~0~ is expected between different populations and the estimated values of R~0~ discussed below are for specific populations in specific environments. The different estimates of R~0~ should not necessarily be interpreted as a range of estimates of the same underlying parameter.
```

alavendelm: ok just changed ref to time since R0 should - by definition - be time-independent

rdvelazquez:
Thanks for the review @alavendelm (that "pending" review feature gets all of us occasionally).

After I accepted your first edit, GitHub marked it as "resolved" and wouldn't let me also accept the second edit so I made that change as a separate commit (https://github.com/greenelab/covid19-review/pull/352/commits/4932687a27d5691cf6d1e664f404e274862f08f0).

PR #376, opened 2020-06-16
Automatically update intro statistics

### LINKED ISSUE #218: Numbers and graph(s) for the covid-19 current status using CI
**Is your feature request related to a problem? Please describe.**

In the intro, at least two items can utilize the power(!) of CI and public databases.

One is the case count, e.g.
At this time, over 92,000 deaths had been reported due to COVID-19 worldwide.
This number should be updated constantly.

The other is (not included currently) a graph such as cumulative case or death count.

**Describe the solution you'd like**

Code bit to pull from public database to fill these two blanks. I am not sure if any of these websites will remain open permanently so I am not sure what the long term solution could be.

**Describe alternatives you've considered**

A database freeze stored in the github repo right before the mx is declared submittable/publishable, which is good for reproducibility but bad for evolving data. Another alternative is to include specific dates for each of these references to clarify the context.

**Additional context**

N/A

agitter:
Updates the introduction so that the death statistics are automatically updated during continuous integration builds.

### Related issues
Closes #218

### Checklist


agitter:
I took initial steps to download the global deaths data from [JHU CSSE](https://github.com/CSSEGISandData/COVID-19/tree/master/csse_covid_19_data/csse_covid_19_time_series) and store some statistics that can be inserted into the manuscript.  A figure is generated as well.

My plan is to update these statistics daily and use the GitHub Actions cached version if they have already been generated for that day.  However, that would mean that only the GitHub Actions version of the manuscript would have the latest statistics and figure.  The AppVeyor version would fall back to the backup versions that I'll commit to the `content` directory.  Is it a problem if the two versions diverge?

Remaining tasks:
- [ ] Check terms of use and attribute appropriately
- [ ] Generate a better figure
- [ ] Set up CI caching to update CSSE statistics at most once per day
- [ ] Format the date
- [ ] Update the static version of the figure and statistics
- [ ] Decide whether GitHub Actions and AppVeyor versions of the manuscript need to be consistent
- [ ] Documentation

cgreene:
What do you think about using a daily scheduled github action in a subrepository and then using the main branch's head instead of a specific commit? https://help.github.com/en/actions/reference/events-that-trigger-workflows#scheduled-events-schedule

cgreene:
This is really interesting. Let me see if I can describe how this works.

In this you would add the actions for the main branch to have a build occur on the `external-resources` branch. The source code for generating the figure and stats all lives on the external resources branch, which has its own conda environment as well.

Does something end up in the main branch as well, or how does the information from the external-resources branch reach the build environment when the document is created? I like where this is going, but I remain just a little bit confused.

agitter:
I haven't attempted anything like this before, so there is still no guarantee that it _does_ work.  However, you've understand my plans correctly.

> In this you would add the actions...

This paragraph is all accurate.  The `external-resources` branch has its own conda environment because it is mostly independent from the main branch.  My vision is that if we want to ingest data from additional external resources, say New York Times or a clinical trials database, those resources would each have their own subdirectories with the respective scripts, stats, and figures.  Their dependencies would be added to the conda environment in the `external-resources` branch.

> Does something end up in the main branch as well

None of the content from the `external-resources` branch will end up being tracked in the main branch.  The content is accessed via URL.  It just so happens that the URL will refer to another branch in this same repository.  Specifically, in this PR I'll update the line in `build.sh`
```
  --template-variables-path=content/csse/csse-stats.json \
```
to something like
```
  --template-variables-path=https://github.com/greenelab/covid19-review/raw/external-resources/csse/csse-stats.json \
```
I haven't tried it yet, but the Manubot docs say that `--template-variables-path` can be a URL to a templates variable file.  The URL we'll use points to the HEAD of the `external-resources` branch.  Therefore, GitHub Actions and AppVeyor builds will always have access to the stats and figure that are updated daily.  The figure is also referenced by URL.  After #382 is merged, we can see whether this works.

In `build.sh` I'll also log which commit was the HEAD of `external-resources` when the build ran.  If we ever wanted to freeze the external resources, we could replace the HEAD by a specific commit in the `--template-variables-path`.

cgreene:
Oh - wow - that's great! It uses github as file storage from the auto-updating branch. Love it! 🤯 ➡️ 😁

agitter:
This is ready for review.  The `external-resources` branch is working as expected.  The statistics and figure show up in the intro.  `build.sh` now logs the `external-resources` commit that was used to generate those figures.

The new scheduled workflow is mostly working but needs to be merged before we can fully test it.  The previous [run failed](https://github.com/greenelab/covid19-review/pull/376/checks?check_run_id=789661338).  This is expected because the `GITHUB_TOKEN` associated with my fork does not have write permission for the `external-resources` branch in `greenelab`.  Downloading new data, generating figures, and committing the updated files all worked as expected.

If we stop building the manuscript regularly, the global deaths figure could get out of sync with the statistics in the text.  The statistics will be updated every time the manuscript is built using the HEAD of the `external-resources` branch at the time of the build.  However, the figure URL will always point to the HEAD of the `external-resources` branch, which will continue to update daily irrespective of whether the manuscript is rebuilt.  I don't think this is an issue.  If it is, I have an idea to fix it in the workflow that updates `external-resources`.

agitter:
@rdvelazquez what you described regarding the HTML and PDF versions going out of sync is correct.  This could also happen with the date and count in the sentence 

> {{csse_deaths}} COVID-19 deaths had been reported worldwide as of {{csse_date_pretty}} (Figure @fig:csse-deaths).

and the actual figure in the HTML version.  The following should be locked once the build is completed:
- Stats and figure in PDF version
- Stats in HTML version

But this will continue updating every day:
- Figure in HTML version

I do think it's possible to make all of that sync.  I'll open a new issue to track that.  But I agree that we should merge first and make sure the basics work.

agitter:
The scheduled build worked correctly:
- [Build log](https://github.com/greenelab/covid19-review/runs/791671916?check_suite_focus=true)
- Commit 75109c1e3d9a473ce0a0f9e539943cab3502664a

PR #382, opened 2020-06-18
External resources

agitter:

This pull request is a partial implementation of the idea in https://github.com/greenelab/covid19-review/pull/376#issuecomment-645328099.  It creates a new branch called `external-resources` that contains automatically updated stats and figures for the manuscript.  I'll add a scheduled workflow in #376 to re-run the scripts in this branch daily.

I'm focusing on the Manubot and workflow aspects of automatically updating the intro global deaths figure.  The figure is intentionally ugly right now.  I'm hoping someone else will volunteer to follow up with another pull request to edit this Python script to make the figure look better.

I created this branch in the `greenelab/covid19-review` repo with the following commands:
```
git checkout --orphan external-resources
git rm -rf .
git commit --allow-empty -m "Initial commit on new branch"
git push origin external-resources
```

### Related issues

This supports #376 and helps address #218.

@rdvelazquez


rdvelazquez:
@agitter This looks like a good. I had the same 🤯 reaction as @cgreene at first but after reading through this as well as https://github.com/greenelab/covid19-review/pull/376 it all seems to generally make sense. Are there some script in #376 that should only be in this branch and not in master?

The complete workflow may be difficult to test without just merging into master (once ready)... I'm ok with doing that and just keeping an eye on it to make sure we fix or revert anything that breaks.

agitter:
> Are there some script in #376 that should only be in this branch and not in master?

Yes, that's correct.  I can remove scripts and other files from #376 once this is merged.

You're right that we'll have to merge and see what happens.  I'll monitor and fix whatever is broken.  Because this PR looks good to you, I'll start by merging this and then work on #376.

PR #403, opened 2020-06-24
Update Author Info - SC

rando2:
@scapone01, I agree with @agitter that you were involved in writing the original draft! I also tried to add the city and country for your affiliation (based on the med school home page), but please feel free to fix this to how you would usually write it!

### DIFF
--- a/content/metadata.yaml
+++ b/content/metadata.yaml
@@ -544,7 +544,7 @@ authors:
       - Writing - Review & Editing
       - Writing - Original Draft
     affiliations:
-      - St. George's University School of Medicine, University Centre, Grenada
+      - St. George's University School of Medicine, St. George's, Grenada
     code of conduct:
       confirmed: yes
     coi:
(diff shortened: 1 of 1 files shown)


content/metadata.yaml
```diff
@@ -542,8 +542,9 @@ authors:
     email: scapone01@gmail.com
     contributions:
       - Writing - Review & Editing
+      - Writing - Original Draft
     affiliations:
-      - St. George's University School of Medicine
+      - St. George's University School of Medicine, University Centre, Grenada
```
scapone01, line 547:
```suggestion
      - St. George's University School of Medicine, St. George's, Grenada
```

scapone01:
> @scapone01, I agree with @agitter that you were involved in writing the original draft! I also tried to add the city and country for your affiliation (based on the med school home page), but please feel free to fix this to how you would usually write it!

@rando2 Thanks for adding that.  I wasn't sure since the full draft was almost completed by the time I jumped in.  I edited my affiliation, so I'm good to go!

PR #409, opened 2020-06-26
Trial Updates to Nutraceuticals

RLordan:
Hi all, just an update to the nutraceuticals section with new trials added. Feel free to edit and review. @rando2 @yemarshall @nilswellhausen 

On another note, I had written a little bit on probiotics as there are trials now being conducted for COVID. However, probiotics are sometimes a little controversial. If anyone has an opinion on that let me know. I have some literature below for anyone interested.  

https://doi.org/10.1080/07391102.2020.1775123
https://doi.org/10.1016/j.virusres.2020.198018
doi:10.1111/LAM.13334
doi:10.3389/fpubh.2020.00186

### Checklist

- [x] Text is formatted so that each sentence is on its own line.

- [x] Pre-prints cited in this pull request have a GitHub issue opened so that they can be reviewed.

PR #410, opened 2020-06-26
Add words to custom dictionary

rando2:
I have been remiss in updating the custom dictionary. This should resolve all of the existing issues prior to the outstanding PRs -- but I used the appveyor output from #409, so it's possible I added a few new words from that PR. 

We (by which I mostly mean "I") should check this every time we merge new text!

### DIFF
--- a/build/assets/custom-dictionary.txt
+++ b/build/assets/custom-dictionary.txt
@@ -62,6 +62,8 @@ axillary
 azithromycin
 BAL
 BALF
+Bansal
+bansalvi
 Bao
 barcoding
 Baric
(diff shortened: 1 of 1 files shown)


build/assets/custom-dictionary.txt
```diff
@@ -735,6 +748,7 @@ O’Meara
 OC
 olfaction
 omics
+opacitiess
```
cbrueffer, line 751: typo?

cbrueffer:
```suggestion
opacities
```

rando2: Yep! Thanks for catching it!

Issue #406, opened 2020-06-25
Incorporate auto-updating clinical trials statistics


agitter:
There are several projects and databases that aggregate and clean COVID-19 clinical trials data.  We could add an auto-updating statistic, figure, and/or table reporting information about clinical trials.

Of those I've reviewed, http://covid19.trialstracker.net/ is appealing.  The dataset is available in a clean csv file, the code is open source at https://github.com/ebmdatalab/covid_trials_tracker-covid, and they are very open to data reuse http://covid19.trialstracker.net/about/.

Does this seem like a good addition?

rando2:
This is incredible! @hufengling and @SiminaB might also be interested conceptually, and I know @nilswellhausen is interested in helping with visualizations.

nilswellhausen:
Sounds like a good idea. Just had a quick look. I think interesting points to illustrate would be the number of trials in any given phase, number of trials for any given study type and maybe a list of the top 10 drugs with the most number of trial. We could then emphasize the point that ressources are not allocated very well (like why do we need 30+ trials for drug x). Although it would be interesting to see which country has the largest amount of trials, I am somewhat hesitant about bringing demographics into this because it could make the impression that country x is better than countries y,z etc.

agitter:
This project's [figures](http://covid19.trialstracker.net/figures/) give a good preview of what the data look like.  @nilswellhausen to your point, we can see there are over 250 trials about hydroxychloroquine.

I'll work on ingesting the data, creating a pandas dataframe with it, and making a proof of concept table or figure.  Once we have the data accessible we can try out more visualizations.

SiminaB:
Looks great! They already include clinicaltrials.gov, which seems to provide most of data, as well as some other registries.
One thing to keep in mind is that a "trial" that's registered in, say, clinicaltrials.gov, is not always what we think of as a "randomized trial" - phase I and II trials are not usually randomized and some of the studies in fact look at symptoms and don't have an intervention (eg "Analysis of clinical characteristics of severe novel coronavirus pneumonia (COVID-19)") - they do have an "intervention" column with some details so that's something we should definitely consider using.

agitter:
[...]

PR #465, opened 2020-07-02
Initial inclusion of COVID-19 TrialsTracker data

agitter:

This pull request add scripts to download, process, and summarize the COVID-19 TrialsTracker data.

### Related issues

#406

@rdvelazquez 

I'll continue working on a simple figure and additional statistics for this initial pull request.  We can discuss what to extract and plot in #406.

The dataset not updated on a daily basis, so I'm tracking two dates.  One is the most recent date of the git commit that updated the data file.  The other is the most recent date of when the trial data was updated, which is the relevant one to report in the manuscript.


rdvelazquez:
This looks really cool @agitter! I pulled your branch down and was able to run it on my machine. I took a quick look at the code and things seem to make sense.

agitter:
I added an initial draft of a figure for the clinical trials data.  I could probably remove some of the uninformative values, like study types with a count of 0 and trials with a "Not Applicable" phase. Any other suggestions for the first version of the figure?

rdvelazquez:
The figures look nice!

rando2:
Very cool @agitter! Is this OK to merge?

agitter:
@rando2 this is ready to merge now if my final updates to the figures look good to you.  I eliminated some of the uninformative categories in the bar graphs.

rando2:
> @rando2 this is ready to merge now if my final updates to the figures look good to you. I eliminated some of the uninformative categories in the bar graphs.

It looks great to me! I also looked through your code and learned how cool the manubot python library is!

SiminaB:
Looks good! I wonder about "Traditional Medicine," which is in first place. Is that the same "Traditional Chinese Medicine" (TCM), which includes interventions like acupuncture etc? I know that sometimes those trials are controversial if they don't have a true placebo/comparator arm.
Are there only 3 recruitment statuses in the data? (Recruiting, Not recruiting, authorised?)

agitter:
I didn't find a definition of "Traditional Medicine" yet, but I believe you are right that it refers to Traditional Chinese Medicine.  I agree those trials can be controversial.  Do you have suggestions on how to handle this?  Should I remove that category from the figure?  Or should I leave it, because that is the true breakdown of the trials data, but we comment on it in the manuscript text?

Yes, those were the only recruitment statuses.  There were another 4 studies with status "No Status Given", which I removed.

SiminaB:
I think we should leave them in, just explain in the text what they refer to and maybe also check if most of them are undertaken in China - otherwise it would seem odd that we don't really discuss TCM as a possible treatment but a plurality of trials involve TCM. I guess trial location might be another interesting thing to pull out, though it may get messy with multi-institution trials.

NickCEBM:
Hey all! Checked this out when I saw the cross-referenced issues. Just to chime in here, Traditional Medicine isn't just TCM, but can also be Traditional Medicines in other contexts. For instance, Iran has a number of herbal (and things like curcumin/honey) interventions being tested as well.

One could probably reasonably disagree with where we've been drawing that distinction but just wanted to let you know It's not just TCM.

SiminaB:
Interesting points @NickCEBM! That's partly why I was suggesting to @agitter to get the location for the trials. Will add that suggestion in https://github.com/greenelab/covid19-review/issues/406

Issue #413, opened 2020-06-28
Note to add more numerical examples of estimated cases vs reported cases or serology-detected vs cases reported


yemarshall:
Towards the end of diagnostics section in the part on Challenges to Diagnostic Approaches / Strategies and Considerations for Determining Whom to Test, there is mention of 1 numerical example of estimated cases versus reported cases. Since this is a variable number, it would be good to add more numerical examples / a range as more data comes out. I would have liked to add this myself, but not quite there yet. Here are two potential examples we could use:
-the 1 referenced example mentioned already in the text right now estimates ~100-fold difference between estimated & reported in late march in Italy.
-In a pdf file titled NY Forward A GUIDE TO REOPENING NEW YORK & BUILDING BACK BETTER May 2020 which I think I got a while back from the ny.gov website or forward.ny.gov website but can't find it again: It says in May there was ~300,000 cases (~1.5% of NY which has a population of ~20x10^6) reported in NY state and that antibody tests showed ~20% positive. So this is a ~10-fold difference between total-based-on-antibody-tests & reported. But I don't have the full information for this data.
-I found an article on Boise, Idaho [doi:10.1128/JCM.00941-20] which just from browsing the article (haven't assessed it yet) looks like they used ~5000 self-selected individuals for antibody tests and found ~2% in late April. But I don't currently have a corresponding reported cases in ratio to population number to compare this to.
-there's probably more data since I last checked (& will be more later)

**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PR #486, opened 2020-07-07
smallAdditionDataExample.diagnostics.md

yemarshall:
I mentioned this a bit back in #413 as a useful type of small addition & am adding it now that I had the chance to find a url I was looking for. Here's the explanation/reason why useful:
Towards the end of diagnostics section in the part on Challenges to Diagnostic Approaches / Strategies and Considerations for Determining Whom to Test, there is mention of 1 numerical example of estimated cases versus reported cases. Since this is a variable number, it would be good to add more numerical examples / a range as more data comes out. I would have liked to add this myself, but not quite there yet. Here are two potential examples we could use:
-the 1 referenced example mentioned already in the text right now estimates ~100-fold difference between estimated & reported in late march in Italy (I didn't check if this estimate is based on serology tests or models or both).
-In a pdf file titled NY Forward A GUIDE TO REOPENING NEW YORK & BUILDING BACK BETTER May 2020 which I think I got a while back from the ny.gov website or forward.ny.gov website but can't find it again: It says in May there was ~300,000 cases (~1.5% of NY which has a population of ~20x10^6) reported in NY state and that antibody tests showed ~20% positive (correction -- ~12%). So this is a ~10-fold difference between total-based-on-antibody-tests & reported. But I don't have the full information for this data.
-I found an article on Boise, Idaho [doi:10.1128/JCM.00941-20] which just from browsing the article (haven't assessed it yet) looks like they used ~5000 self-selected individuals for antibody tests and found ~2% in late April. But I don't currently have a corresponding reported cases in ratio to population number to compare this to.
-there's probably more data since I last checked (& will be more later)

@rdvelazquez @agitter & whoever wrote this part (will need to check) or anyone else

### Checklist


yemarshall:
question for @agitter , I put the simpler url above of
ny.governor.gov/nyforwardbook
but there's also another url which can work
https://www.governor.ny.gov/sites/governor.ny.gov/files/atoms/files/NYForwardReopeningGuide.pdf
Would you double-check that the url I put is okay? if not, we can switch to the other one.
thanks!

content/10.diagnostics.md
```diff
@@ -173,4 +173,4 @@ Specifically, the individuals at risk include:
 
 However, this method of testing administration does not detect a high proportion of infections and does not allow for test-and-trace methods to be used.
 Individuals who are asymptomatic (i.e. potential spreaders) and individuals who are able to recover at home are therefore often unaware of their status.
-For instance, a recent study from the Imperial College estimates that in Italy the true number of infections is around 5.9 million against the 70,000 detected as of March 28th [@doi:10.25561/77731].
+For instance, a recent study from the Imperial College estimates that in Italy the true number of infections is around 5.9 million (total population ~60 million) against the 70,000 detected as of March 28th [@doi:10.25561/77731]; another example about New York state mentions for May 2020 ~300,000 cases reported (total population ~20 million) or ~1.5% of the population while antibody tests show ~20% positive [@url:ny.governor.gov/nyforwardbook].
```
rando2, line 176:
```suggestion
For instance, a recent study from Imperial College estimates that in Italy the true number of infections is around 5.9 million in a total population ~60 million, compared to the 70,000 detected as of March 28th [@doi:10.25561/77731].
Another analysis, which examined New York state, indicated that as of May 2020, approximately 300,000 cases had been reported in a total population ~20 million, corresponding to ~1.5% of the population, but up to ~20% of the population tested positive when evaluated with antibody tests [@url:ny.governor.gov/nyforwardbook].
```
I'm not sure if this is what you were saying but I would recommend splitting up into two sentences just because it's so much info! I actually can't get the NYForwardBook url to work. In the [version I found](https://www.governor.ny.gov/sites/governor.ny.gov/files/atoms/files/NYForwardReopeningGuide.pdf) it sounds like NYC is ~20% but the rest of the state is lower?

yemarshall: good catch! -- I had double-checked but I guess I should have triple checked! So, in addition to your suggested changes above we need to change the last number from ~20% to ~12%.
-so your changes work in how your rewrote the sentence
plus
-we need to change ~20% to ~12%
plus
-we should change the url to https://www.governor.ny.gov/sites/governor.ny.gov/files/atoms/files/NYForwardReopeningGuide.pdf
(I guessed this might be necessary)

yemarshall:
@rando2 good catch! -- I had double-checked but I guess I should have triple checked! So, in addition to your suggested changes above we need to change the last number from ~20% to ~12%.
-so your changes work in how your rewrote the sentence & I added that rewrite already
Now, we still need to:
-change ~20% to ~12%
plus
-change the url to https://www.governor.ny.gov/sites/governor.ny.gov/files/atoms/files/NYForwardReopeningGuide.pdf
(I guessed this might be necessary)

rando2:
> @rando2 good catch! -- I had double-checked but I guess I should have triple checked! So, in addition to your suggested changes above we need to change the last number from ~20% to ~12%.
> -so your changes work in how your rewrote the sentence & I added that rewrite already
> Now, we still need to:
> -change ~20% to ~12%; also, I just noticed something in your rewrite -- I'm not sure if "up to ~12%" makes sense (I can guess biology reasons for ~raw/~less-processed antibody test numbers to be an overestimate or an underestimate & I've only read a bit about some serology tests & don't know if the ~12% accounts for possible overestimate or underestimate / what type of approximation it is); so we can keep your rewrite but remove the two words "up to" as well as changing ~20% to ~12%;
> plus
> -change the url to https://www.governor.ny.gov/sites/governor.ny.gov/files/atoms/files/NYForwardReopeningGuide.pdf
> (I guessed this might be necessary)

Oh, all I meant by "up to" was that in NYC it was extremely high, as you note! So you could provide the range to indicate that it could actually be a very significant chunk of the population, depending on region. But it's up to you how to handle it, I think either is fair!

content/10.diagnostics.md
```diff
@@ -173,4 +173,5 @@ Specifically, the individuals at risk include:
 
 However, this method of testing administration does not detect a high proportion of infections and does not allow for test-and-trace methods to be used.
 Individuals who are asymptomatic (i.e. potential spreaders) and individuals who are able to recover at home are therefore often unaware of their status.
-For instance, a recent study from the Imperial College estimates that in Italy the true number of infections is around 5.9 million against the 70,000 detected as of March 28th [@doi:10.25561/77731].
+For instance, a recent study from Imperial College estimates that in Italy the true number of infections is around 5.9 million in a total population ~60 million, compared to the 70,000 detected as of March 28th [@doi:10.25561/77731].
+Another analysis, which examined New York state, indicated that as of May 2020, approximately 300,000 cases had been reported in a total population ~20 million, corresponding to ~1.5% of the population, but up to ~20% of the population tested positive when evaluated with antibody tests [@url:ny.governor.gov/nyforwardbook].
```
rando2, line 177:
```suggestion
Another analysis, which examined New York state, indicated that as of May 2020, approximately 300,000 cases had been reported in a total population of approximately 20 million, corresponding to ~1.5% of the population, but 12% of test individuals statewide and up to ~20% of the population in some places tested positive when evaluated with antibody tests [@url:https://www.governor.ny.gov/sites/governor.ny.gov/files/atoms/files/NYForwardReopeningGuide.pdf].
```

rando2:
@yemarshall if you don't object, I'll go ahead and merge it! We can always refine later.

content/10.diagnostics.md
yemarshall: @rando2 to make the last clause slightly clearer, maybe rechange the order to: "but when evaluated with antibody tests ~12% of individuals were positive statewide and up to ~20% in some places were positive" and then the correct url reference. (reason for clarifying: right now the modified clause is ambiguous with respect to what the ~12% means exactly)

yemarshall:
Could try to make the interpretation of the numbers easier towards the end:
"but ~12% of individuals tested statewide and up to ~20% in some places tested positive when evaluated with antibody tests"
change to
"but when evaluated with antibody tests ~12% of individuals statewide and up to ~20% in some places tested positive"
OR just remove the first "tested" so reads instead
"but ~12% of individuals statewide and up to ~20% in some places tested positive when evaluated with antibody tests"

yemarshall:
See PR #571 to complete fixing this (we got busy & delayed it to later).

PR #519, opened 2020-07-20
Remove typos from custom dictionary

rando2:
I noticed the spellchecker wasn't catching some typos -- turns out, some typos had made it in! I didn't check all of the acronyms, but I used Microsoft Word's spellchecker + Google to try to identify any words that shouldn't be in here. I also removed British English spellings. I'm not sure whether capitalization matters but I tried to make sure caps were present where needed (e.g., brand names or software), although I likely didn't catch all of these.

When this builds, I will add in fixes for anything it picks up in the text.


agitter:
> I'm not sure whether capitalization matters

Currently spell checking is case-insensitive.  @cgreene showed me how to make it case-sensitive, and I've been meaning to push that into manubot/rootstock.

rando2:
@agitter I was pretty sure it wasn't case sensitive, but figured it wouldn't hurt to try to set us up for backwards compatibility down the road in case that changed, so sounds like we're kind of on the same page!

PR #545, opened 2020-07-27
Attempt to count TCZ trials

rando2:

@agitter I was hoping to tabulate the number of tocilizumab trials (and then potentially do this for some of the other therapeutics we highlight in the Therapeutics section as well). I would love to know if you think this is a reasonable approach to exporting the count from the EBM data -- and apologies if I've totally missed the mark! If this is reasonable, I was hoping to add in a reference to the tocilizumab_ct variable in the text. Thank you so much for your help in getting acquainted with CI!

@agitter or @rdvelazquez 

### Checklist

### DIFF
--- a/ebmdatalab/generate-ebmdatalab-stats.py
+++ b/ebmdatalab/generate-ebmdatalab-stats.py
@@ -124,7 +124,7 @@ def main(args):
         f'https://github.com/greenelab/covid19-review/raw/$FIGURE_COMMIT_SHA/{args.output_figure}.svg'
     
     # Tabulate number of trials for pharmaceuticals of interest
-    ebm_stats['ebm_tocilizumab_ct'] = trials_df['intervention'].str.contains('tocilizumab', case=False).sum()
+    ebm_stats['ebm_tocilizumab_ct'] = str(trials_df['intervention'].str.contains('tocilizumab', case=False).sum())
     
     with open(args.output_json, 'w') as out_file:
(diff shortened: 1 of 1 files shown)


ebmdatalab/generate-ebmdatalab-stats.py
```diff
@@ -123,6 +123,9 @@ def main(args):
     ebm_stats['ebm_trials_figure'] = \
         f'https://github.com/greenelab/covid19-review/raw/$FIGURE_COMMIT_SHA/{args.output_figure}.svg'
     
+    # Tabulate number of trials for pharmaceuticals of interest
+    ebm_stats['tocilizumab_ct'] = trials_df['intervention'].str.contains('tocilizumab', case=False).sum()
```
agitter, line 127:
```suggestion
    ebm_stats['ebm_tocilizumab_ct'] = trials_df['intervention'].str.contains('tocilizumab', case=False).sum()
```

Your approach is correct, and this should work.  I'm using the convention where the variable prefix (`ebm_`) denotes the data source to help with later debugging if needed.

rando2:
@agitter I'm getting 68 this morning, which sounds pretty close! And that is a really important note, thank you so much!

agitter:
@rando2 your update worked.  The last nightly build added `"ebm_tocilizumab_ct": "68",` to `ebmdatalab-stats.json`

rando2:
Yay! Thank you so much for checking and for the help, @agitter!

PR #553, opened 2020-07-30
Add results of recent mRNA-1273 study

cgreene:
### Description of the proposed additions or changes

I updated the text to include the results of [@doi:10.1056/NEJMoa2022483]

### Related issues

#556 

### Suggested reviewers (optional)

@likhithakolla @esell17 @rmrussell @nilswellhausen 

### Checklist

- [x] Text is formatted so that each sentence is on its own line.

- [ ] Pre-prints cited in this pull request have a GitHub issue opened so that they can be reviewed.

PR #554, opened 2020-07-30
remove trailing ws from 20.treatments.md

rando2:
Remove trailing white space from the therapeutics section


rando2:
If this looks ok, we can rebase #553 to make it more clear what text @cgreene edited!

rando2:
Thanks @agitter!

PR #544, opened 2020-07-27
Fix Manubot citations in intro, pathogenesis, diagnostics

agitter:
### Description of the proposed additions or changes

This corrects broken references for the first sections of the review.  I made my best attempt to specify authors for websites.  Here are a few unresolved questions:
- #396 added `doi:10.1101/2020.05.14.09613`, and I can't figure out what bioRxiv preprint that sound be.  Any ideas @rando2 or @ajlee21?  It was added to line 108.
- Is Cellex the author of https://www.fda.gov/media/136625/download ?
- Who should be listed as the author of the NY Forward guide (https://www.governor.ny.gov/sites/governor.ny.gov/files/atoms/files/NYForwardReopeningGuide.pdf)?  I didn't include one.  Maybe it is okay to leave it empty.

### Related issues

#371

### Checklist

- [X] Text is formatted so that each sentence is on its own line.

- [X] Pre-prints cited in this pull request have a GitHub issue opened so that they can be reviewed.

PR #560, opened 2020-07-31
Fix Manubot citations in Therapeutics and Appendix 1

### LINKED ISSUE #371: Review all references and make manual updates
Before the first journal submission, I will go through the references to check completeness and look for obvious errors.  I can add manual references for anything missing metadata.  I've noticed this in some URL citations.

I'll wait until most of the content has stabilized before reviewing.

agitter:

Manual references for Manubot citations.  I wasn't sure what to use for the author for the EU clinical trial https://www.clinicaltrialsregister.eu/ctr-search/trial/2020-000936-23/FR  @rdvelazquez what would be most consistent with how you handled clinicaltrials.gov trials?

### Related issues

Closes #371
Follows up on #544

### Checklist


rdvelazquez:
The [Zotero translator for clinicaltrials.gov](https://github.com/zotero/translators/blob/master/clinicaltrials.gov.js) has a `creators` field and lists both the lead investigator (as `author`) and the responsible party (as `contributor`) unless there is only a responsible party listed, in which case the responsible party is considered the `author`.

What you have looks good it may just be worth double checking how the EU trial citation looks compared to the us (.gov) one once they are formatted.

rdvelazquez:
This is how it's showing up on the build that I see from appVeyor:
Screen Shot 2020-07-31 at 10 24 14 AM

agitter:
There was a UTF-8 encoding error, so my manual references weren't being used.  Hopefully that's fixed in the next build.

For this EU trial, I didn't see a lead investigator or responsible party.  That probably means we should exclude the author field.

agitter:
Here's how the EU trial now appears.  Looks good to me.
image

I caught two more errors while reviewing the AppVeyor build.  The other manual references worked.

agitter:
I finally fixed the last reference that was causing problems because of `---` in the URL:
image

As a note for my future self, when using a `raw:` citation, I need to remove the `note` field from the JSON file.  Otherwise Manubot reads the `id` from the note text and overrides the `id` field in the JSON.

PR #552, opened 2020-07-30
Discussion: inital draft of equity concerns (work in progress)

rando2:
### Description of the proposed additions or changes

This is @tlukan's initial draft of the discussion! He is an undergraduate contributor who is interested in healthcare inequity and how that can affect research on COVID-19.
(The PR says "@rando2 wants to merge", but I just helped get the PR set up with @tlukan's text)

This is a work in progress so right now it's better to focus on the ideas that @tlukan proposed rather than specifics of the writing

### Related issues

#460 

### Checklist

- [x] Text is formatted so that each sentence is on its own line.

- [ ] Pre-prints cited in this pull request have a GitHub issue opened so that they can be reviewed.

PR #569, opened 2020-08-05
Map where clinical trials are recruiting

rando2:

This is a proposed addition to the EBM Data Lab COVID-19 TrialsTracker analysis that is used to generate some of the existing figures. Here, the countries represented in the EBM dataset are tabulated, with single-country clinical trials and multi-country clinical trials counted separately. Tabulation is done based on ISO codes, which are easier to match than country names. 

Here is what it creates:
ebmdatalab-map

Questions:
- Is modifying the environment.yml file sufficient to install the new packages on the virtual environment? This seems too simple, but I haven't figured out where else I need to add them. 

Concerns:
- There is some data cleaning that has to happen to match the EBM dataset to ISO codes -- I'm not sure what the best way to represent this information for reproducibility is (for example, if you search for South Korea, you'll see where I hard-coded in the ISO code based on the official name). I did these automatically with pycountry except where I could not find any workaround ("South Korea" and "Democratic Republic of [the] Congo")
- For whatever reason it's very hard to handle side-by-side choropleths where the scales are quite different. I don't know what you think of these two color scales and the off-white background. 70 seems to be approximately equal to purple in the top figure...
- I did run a linter, but some of its advice seemed a little counterintuitive, so I'm happy to hear if there's anything formatted in a confusing way.

I'm happy to hear suggestions either on the code or visualization side! 

### Related issues

#552

@tlukan @agitter @cgreene @rdvelazquez

### DIFF
--- a/ebmdatalab/generate-ebmdatalab-stats.py
+++ b/ebmdatalab/generate-ebmdatalab-stats.py
@@ -37,6 +37,45 @@ def extract_citekey(results_url):
         citekey = short_doi_url.replace('https://doi.org', 'doi:10')
     return citekey
 
+def assign_ISO(countries):
+    # Match country names with ISO codes
+    # Input: pd.Series of country names
+    # Returns: dictionary of matches
+
+    # Need to hard code a few countries that aren't registered using standard names, so
+    # initializing the single_country_codes database with these values
+    country_codes = {"South Korea": "KOR", "Democratic Republic of Congo": "COD",
+                     "Democratic Republic of the Congo": "COD"}
+
+    # Identify the most likely 3-letter ISO code for each country
+    failed_matches = list()
+    for country in countries:
+        if country not in country_codes.keys():
+            try:
+                hit = pycountry.countries.get(name=country)
+                if hit == None:
+                    # If the name isn't an exact match, try alternatives
+                    # .search_fuzzy matching returns a list, whereas .get retrieves data as class Country
+                    hit = pycountry.countries.search_fuzzy(country)
+                    if len(hit) > 1:
+                        hit = pycountry.countries.search_fuzzy(country + ",")
+                    elif type(hit) == None:
+                        hit = pycountry.countries.get(official_name=country)
+            except LookupError:
+                failed_matches.append(country)
+                continue
+
+            if type(hit) == list and len(hit) == 1:
+                country_codes[country] = hit[0].alpha_3
+            elif type(hit) == list or type(hit) == None:
+                failed_matches.append(country)
+            else:
+                country_codes[country] = hit.alpha_3
+
+    # Print warning about failures and return successes as dictionary
+    print("Failed to assign country codes to:", ", ".join(failed_matches))
+    return(country_codes)
+
 # Inspired by https://github.com/greenelab/meta-review/blob/master/analyses/deep-review-contrib/03.contrib-stats.ipynb
 def main(args):
     '''Extract statistics from the EBM Data Lab COVID-19 TrialsTracker dataset'''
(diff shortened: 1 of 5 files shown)


ebmdatalab/generate-ebmdatalab-stats.py
```diff
@@ -1,10 +1,13 @@
 import argparse
 import datetime
 import json
+from datetime import date
 import matplotlib.pyplot as plt
 import os
 import pandas as pd
 import urllib.request
```
rando2, line 8: @agitter the linter was very insistent that we don't actually use urllib anywhere.

cgreene:
🇫🇷 is missing?

ebmdatalab/generate-ebmdatalab-stats.py
agitter: Thanks for checking.  You can delete `urllib.request`.  My initial draft needed it for the short DOI functionality, but then I realized I could import `get_short_doi_url` from Manubot instead.

agitter:
> Is modifying the environment.yml file sufficient to install the new packages on the virtual environment

Yes, that's all you need to do.  This GitHub actions step magically takes care of everything else: https://github.com/greenelab/covid19-review/blob/e3bd8788f136f10b3f5ea117b15e48049c6a96a2/.github/workflows/update-external-resources.yaml#L17-L23

rando2:
Per @cgreene's observation, something is very wrong with France (and Norway, and a few other places) in this geopandas dataset: https://github.com/geopandas/geopandas/issues/1041
I will add manual fixes for these and upate!

rando2:
Vive La France!
ebmdatalab-map

RLordan:
I have zero technical knowledge to help... but looks really cool folks, thank you

rando2:
Per @tlukan's suggestion, here is a description for what these figures show (possible draft for a legend):
The EBM Data Lab COVID-19 TrialsTracker dataset was used to identify where, at the country level, each clinical trial is or was enrolling participants. The number of clinical trials operating in each country was tabulated and is shown here through the color overlay on the map. The top figure shows the number of clinical trials specific to each country (e.g., trials recruiting only in that country). The bottom figure shows the same but for countries recruiting in multiple countries. Note that the scale differs between the top and bottom plots.

rando2:
Currently this is missing the following ISO codes: 'LIE', 'MLT', 'GUF', 'SMR', 'HKG', 'GIB', 'BHR', 'MCO', 'UMI', 'IMN', 'SGP', 'MTQ'
In most cases, I think the geopandas datasets lumps these in with other countries for various political reasons (e.g., HKG is Hong Kong, GUF is French Guiana). I'm working on splitting the maps to handle them separately because culturally they seem worth analyzing separately.

ebmdatalab/generate-ebmdatalab-stats.py
```diff
@@ -35,6 +37,64 @@ def extract_citekey(results_url):
...
+                if isinstance(hit, type(None)):
+                    # .get retrieves data as class Country
+                    # if it can't find a match, try alternative methods
+                    # .search_fuzzy matching returns a list
+                    hit = pycountry.countries.search_fuzzy(country)
+                    if len(hit) > 1:
+                        hit = pycountry.countries.search_fuzzy(country + ",")
+                    elif isinstance(hit, type(None)):
+                        hit = pycountry.countries.get(official_name=country)
+            except LookupError:
```
cgreene, line 68: There are two get calls here - does it matter which one raises the lookup error? I'm imagining that search_fuzzy wouldn't raise one form its name.

rando2: It is only the `pycountry.countries.get()` command that throws this error, `.search_fuzzy()` I believe returns an empty list. Should this section be reformatted to better reflect that? I wasn't quite sure the best way to handle this many levels of possible alternative strategies, although in retrospect I could just make a function that would return a value whenever it first succeeds.

cgreene: I noticed there are two .get calls, and if you hit this error the first one you will never do the fuzzy search. We also won't know which .get call failed. I'm not sure that either of these matter, so I wanted to raise them to see if it would make a difference.

rando2:
Thank you so much @cgreene! I believe this handles exceptions a little more consistently. I checked with @mprobson and we couldn't brainstorm a more elegant structure, though there could be something we're not thinking of!

ebmdatalab/generate-ebmdatalab-stats.py
```diff
@@ -35,6 +37,93 @@ def extract_citekey(results_url):
...
+            # few reasons. Use function to avoid LookupError issues and
+            # try all the different ways that might help to match a
+            # country name to its ISO code
+            hit = find_country(country)
+            if not isinstance(hit, type(None)):
+                country_codes[country] = hit.alpha_3
+            else:
+                failed_matches.append(country)
+    # Print warning about failures and return successes as dictionary
+    print("Could not assign country codes to:", ", ".join(failed_matches))
```
agitter, line 107:
```suggestion
    print("Could not assign country codes to: ", ", ".join(failed_matches))
```
Add whitespace here?

ebmdatalab/generate-ebmdatalab-stats.py
```diff
@@ -97,35 +186,104 @@ def main(args):
...
+    # Remove any leading/trailing whitespace that may result from splitting above
+    unique_countries = single_countries.append(multi_countries).str.strip().drop_duplicates()
+    country_codes = pd.DataFrame.from_dict(assign_ISO(unique_countries),
+                                           orient="index",
+                                           columns=["iso_a3"])
+
+    # Map the ISO codes onto the country data and count the frequency
+    single_countries_codes = pd.DataFrame(single_countries,
+                                          index=single_countries).join(country_codes)["iso_a3"]
+    single_countries_codes = single_countries_codes.dropna()
```
agitter, line 233: You're using the 3 letter country code here, so it looks like we won't have a problem with Namibia (2 letter code `NA`), right?

h/t https://twitter.com/trang1618/status/1281339098956009472

ebmdatalab/generate-ebmdatalab-stats.py
```diff
@@ -97,35 +186,104 @@ def main(args):
...
+    ax2.annotate(f'Source: EBM Data Lab COVID-19 TrialsTracker, %s' %
+                 date.today().strftime("%b-%d-%Y"),
+                 xy=(0, 0), xycoords="axes points")
+
+    plt.savefig(args.output_map + '.png', bbox_inches="tight")
+    plt.savefig(args.output_map + '.svg', bbox_inches="tight")
+
+    print(f'Wrote {args.output_map}.png and {args.output_map}.svg')
+
     # The placeholder will be replaced by the actual SHA-1 hash in separate
```
agitter, line 278: I suggest moving this block to follow line 208 (the `Wrote {args.output_figure}.png and {args.output_figure}.svg` message) to keep the trials figure code grouped together.

ebmdatalab/generate-ebmdatalab-stats.py
```diff
@@ -97,35 +186,104 @@ def main(args):
...
     # script after the updated image is committed
     ebm_stats['ebm_trials_figure'] = \
         f'https://github.com/greenelab/covid19-review/raw/$FIGURE_COMMIT_SHA/{args.output_figure}.svg'
-    
+
```
agitter, line 282:
```suggestion
ebm_stats['ebm_map_figure'] = \
        f'https://github.com/greenelab/covid19-review/raw/$FIGURE_COMMIT_SHA/{args.output_map}.svg'
```

To save the figure URL for later use in the manuscript, you should be able to do the above.  `$FIGURE_COMMIT_SHA` will be replaced when the `version-figures.sh` script on the external-resources branch runs.

ebmdatalab/generate-ebmdatalab-stats.py
rando2: Oh wow!!!! I think it should be ok, it looks like alpha_3='NAM'. But that is a really good issue to be aware of 😅

rando2:
Thank you both so much for the review!

PR #571, opened 2020-08-06
finishFixClarificationInFixSuggestion.diagnostics.md

yemarshall:
In PR #486 I added a sentence for the reasons described there, but another contributor noticed I had made a typo with "20%" instead of "12%" (thanks!) -- this was partially corrected but along with a great added point which needed slightly more clarity and then we got busy and didn't get to them. So I'm fixing it now as it is an important added point (thanks again!) and by now I have briefly read about even more indications of studies/data at higher resolutions which makes the point even more interesting. What I have now is mathematically & biologically & grammatically correct. Basically, it was suggested also to indicate that different regions/sub-regions had different numbers which was a great suggestion but we didn't finish incorporating it fully correctly -- if you say "up to" or "above" etc. based on this kind of sampling data you are only referring to the specific data and resolution but not what it means about the population at various possible resolutions, so instead I just added a parentheses which explains the general suggested added point.
This description is long but the fix is short! This was just delayed from earlier in PR#486 & it should be a quick fix -- if it gets merged as is then that works, another quick correct option is to use the sentence fix but with the parenthetical clause removed and then merge, otherwise there's no time allocated for a conversation about this.

### Checklist


content/10.diagnostics.md
```diff
@@ -174,4 +174,4 @@ Specifically, the individuals at risk include:
...
 For instance, a recent study from Imperial College estimates that in Italy the true number of infections is around 5.9 million in a total population ~60 million, compared to the 70,000 detected as of March 28th [@doi:10.25561/77731].
-Another analysis, which examined New York state, indicated that as of May 2020, approximately 300,000 cases had been reported in a total population of approximately 20 million, corresponding to ~1.5% of the population, but 12% of individuals tested statewide and up to ~20% in some places tested positive when evaluated with antibody tests [@url:https://www.governor.ny.gov/sites/governor.ny.gov/files/atoms/files/NYForwardReopeningGuide.pdf].
+Another analysis, which examined New York state, indicated that as of May 2020, approximately 300,000 cases had been reported in a total population of approximately 20 million, corresponding to ~1.5% of the population, but 12% of individuals sampled statewide were estimated as positive through antibody tests (along with indications of spatial heterogeneity at higher resolution) [@url:https://www.governor.ny.gov/sites/governor.ny.gov/files/atoms/files/NYForwardReopeningGuide.pdf].
```
yemarshall, line 177:
```suggestion
Another analysis, which examined New York state, indicated that as of May 2020, approximately 300,000 cases had been reported in a total population of approximately 20 million, corresponding to ~1.5% of the population, but ~12% of individuals sampled statewide were estimated as positive through antibody tests (along with indications of spatial heterogeneity at higher resolution) [@url:https://www.governor.ny.gov/sites/governor.ny.gov/files/atoms/files/NYForwardReopeningGuide.pdf].
```

PR #408, opened 2020-06-25
For Underexplored Therapeutics/Prophylactics: Interactions with Innate Immune Response (TLR agonists/antagonists & BCG trained immunity)

yemarshall:
I put in a paragraph on another type of possible partial-therapeutic/partial-prophylactic that is under clinical trials and also related possible prophylactics under research -- things that interact with innate immune response, mainly TLR agonists/antagonists as drafted by @rando2 at end of #349 & BCG for trained immunity clinical trials drafted by me and with feedback from @rando2 and @rlordan incorporated.
Note also, if review please also consider suggestion to expand section title of "Underexplored Therapeutics" to "Underexplored Therapeutics and Prophylactics"?

### Suggested reviewers (optional)

@rando2 @RLordan @matfax @gregszetoAI @nilswellhausen @ajlee21 

### Checklist

- [ ] Text is formatted so that each sentence is on its own line.

- [ ] Pre-prints cited in this pull request have a GitHub issue opened so that they can be reviewed.

PR #349, opened 2020-05-27
Additional mechanisms for how SARS-CoV-2 bypasses immune response

rando2:
### Description of the proposed additions or changes
This is a first-pass effort to add immunological background. I'm trying to go back through the intro and add in all the information that I have been needing to review in order to understand the pathogenesis, diagnostics, and therapeutics sections. This will be incredibly basic to anyone whose background is actually in immunology, but I think it's important to include the basics because they are so pivotal to understanding the rest of the text, and the audience for this review is a general scientific audience. 

This is the first of at least 3 PRs, as I have been working on content about the adaptive immune response and the cytokine storm as well! After the text is merged, the plan is then to go through the text as a whole and make sure all of these ideas connect coherently (e.g., terminology, aggregate all the cytokine storm explanations in one place)

**Please read skeptically and correct this text as needed!** I had to spend a long time reviewing the terminology, and I'm sure there are still issues.

### Related issues

#233 #335 

### Suggested reviewers (optional)

@nilswellhausen @RLordan @yemarshall @ajlee21 

### Checklist

- [x] Text is formatted so that each sentence is on its own line.

- [ ] Pre-prints cited in this pull request have a GitHub issue opened so that they can be reviewed.

PR #578, opened 2020-08-09
possibleQuickFixPossibleTypoUnderexplored.treatments.md

yemarshall:
#408 Combined a couple of related ideas together.  I tried to fill in context for the ideas but I may not have filled in enough context to include both TLR agonists and TLR antagonists. A quick fix may be to just remove the word "antagonist" because we don't have references for it currently (right now the reference shows some evidence for possible prophylaxis with TLR agonist in aged mice but does not discuss antagonist, so I realized later it sounds funny to include antagonist without explaining more the idea or showing specific references).
Quick fix 1:
The article referenced only speaks of receptor agonists. There's no reference or context given for receptor antagonist here, so makes sense to just remove word. It's possible that what was meant was a more complex strategy, but there's no context or research referenced for it so simpler to remove.
Quick fix 2:
Also, quick fix (2 words changed) there was a choice of LPS as an example but based on the reference maybe makes more sense to use CpG as an example for following reason:
In the particular reference used, 4 agonists were investigated & only poly(I:C) was investigated a lot and showed the most effect. The 2nd most effect was seen with CpG in the small data provided. The least effect was seen with LPS in small data provided. I think maybe the idea of mentioning LPS was because LPS may have a different net relation/effect on immune system as described for example in another reference (the review on trained immunity) but that's not discussed in the particular reference for this sentence and we don't discuss LPS elsewhere in this paragraph. Again, this is an interesting idea but we don't have references for it so maybe better to just include the 1 agonist that was most investigated and then the agonist that showed the 2nd most effect.
These are just quick suggestions/possibilities. (If we don't include them, then I think the general idea may still be understood, but worth at least removing "antagonist" because it's sort of confusing with just the 1 reference that focuses on the idea and examples of "agonist")

### Checklist


rando2:
I actually don't think this is a typo-- it's meant to say that broadly disrupting TLR function might be worth exploring! For example, here is an article that uses both terms:

> Investigations of TLR agonists have also drawn interest into the prospective use of TLR antagonists to decrease inflammatory responses mediated by the innate immune system.

https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2832843/

Thank you for double-checking, though, always good to be sure with terminology!

content/20.treatments.md
```diff
@@ -941,7 +941,7 @@ There is also some research into possible potential therapeutics or prophylactic
 For example, there are a variety of toll-like receptors (TLRs) which are examples of pattern recognition receptors (PRRs), innate immune response components which recognize pathogen-associated molecular patterns (PAMPs) and damage-associated molecular patterns (DAMPs).
 Note that TLRs form a part of innate immune recognition and can more generally contribute to promoting both innate and adaptive responses [@ISBN:9780815332183].
 In mouse models, poly(I:C) and lipopolysaccharide, which are agonists of toll-like receptors TLR3 and TLR4, respectively, showed protective effects when administered prior to SARS-CoV exposure [@doi:10.1128/JVI.01410-12].
```
yemarshall, line 943:
```suggestion
In mouse models, poly(I:C) and CpG, which are agonists of toll-like receptors TLR3 and TLR9, respectively, showed protective effects when administered prior to SARS-CoV exposure [@doi:10.1128/JVI.01410-12].
```

yemarshall: In the particular reference used, 4 agonists were investigated & only poly(I:C) was investigated a lot and showed the most effect. The 2nd most effect was seen with CpG in the small data provided. The least effect was seen with LPS in small data provided. I think maybe the idea of mentioning LPS was because LPS may have a different net relation/effect on immune system as described for example in another reference (the review on trained immunity) but that's not discussed in the particular reference for this sentence and we don't discuss LPS elsewhere in this paragraph.

yemarshall:
I understood the idea of net effect but the term agonist/antagonist next to a receptor like TLRx suggests the direct relationship so I thought it makes sense to write the same term as the reference (since the reference has more experience on these topics than me).

yemarshall:
I understood the idea of net effect but the term agonist/antagonist next to a receptor like TLRx suggests the direct relationship so I thought it makes sense to write the same term as the reference.

yemarshall:
I understood the idea of net effect but the term agonist/antagonist next to a receptor like TLRx suggests the direct relationship so I thought it makes sense to write the same term as the reference cited. I was giving a quick fix as explained in the PR description because a more complex strategy / other ideas were not described or specifically referenced. If you'd like both then a quick fix could be ...agonists (and possibly antagonists)... but that still doesn't work great unless there's reference. Further, if you talk about antagonist or of reducing immune response then you'd need to specify if you're still talking about prophylaxis or have switched to later stage treatment. I just put in a quick simple fix so that what's there is just what's referenced and described and for which there's already at least some experimental data and strategy pointed out in the text.

yemarshall:
see edited/updated description, thanks.

content/20.treatments.md
```diff
@@ -940,8 +940,8 @@ Upcoming work may explore the specificity of nAbs for their target, mechanisms b
 There is also some research into possible potential therapeutics or prophylactics which interact with components of the innate immune response.
 For example, there are a variety of toll-like receptors (TLRs) which are examples of pattern recognition receptors (PRRs), innate immune response components which recognize pathogen-associated molecular patterns (PAMPs) and damage-associated molecular patterns (DAMPs).
 Note that TLRs form a part of innate immune recognition and can more generally contribute to promoting both innate and adaptive responses [@ISBN:9780815332183].
-In mouse models, poly(I:C) and lipopolysaccharide, which are agonists of toll-like receptors TLR3 and TLR4, respectively, showed protective effects when administered prior to SARS-CoV exposure [@doi:10.1128/JVI.01410-12].
```
yemarshall, line 943: change word exposure to infection (more biologically precise). i.e., should be for clarity and specificity to reference: "...prior to SARS-CoV infection [..."
again, this sentence and the next were part of the draft of #349 for edits (the 2nd file) and we decided later to combine it with other similar ideas into #408. It was a great article find, but since it was just a draft requires a bit of fixing so that that it corresponds better to the reference. (Also if we keep "antagonists" in next sentence without explaining then it might superficially sound like either we are not using the prophylaxis in the reference or that we are precluding effective use of the prophylaxis in the reference, unless it's part of a more complex strategy which needs to be described; in order to explain why "antagonist" might be something worth researching we'd need more references and description.)

content/20.treatments.md
```diff
@@ -940,8 +940,8 @@ Upcoming work may explore the specificity of nAbs for their target, mechanisms b
...
-Therefore, TLR agonists and antagonists hold some potential for broad-spectrum prophylaxis.
+In mouse models, poly(I:C) and CpG, which are agonists of toll-like receptors TLR3 and TLR9, respectively, showed protective effects when administered prior to SARS-CoV exposure [@doi:10.1128/JVI.01410-12].
```
yemarshall, line 943:
```suggestion
In mouse models, poly(I:C) and CpG, which are agonists of toll-like receptors TLR3 and TLR9, respectively, showed protective effects when administered prior to SARS-CoV infection [@doi:10.1128/JVI.01410-12].
```
change word exposure to infection (more biologically precise). i.e., should be for clarity and specificity to reference: "...prior to SARS-CoV infection [..."
again, this sentence and the next were part of the draft of #349 for edits (the 2nd file) and we decided later to combine it with other similar ideas into #408. It was a great article find, but since it was just a draft requires a bit of fixing so that that it corresponds better to the reference. (Also if we keep "antagonists" in next sentence without explaining then it might superficially sound like either we are not using the prophylaxis in the reference or that we are precluding effective use of the prophylaxis in the reference, unless it's part of a more complex strategy which needs to be described; in order to explain why "antagonist" might be something worth researching we'd need more references and description.)

PR #591, opened 2020-08-10
Possible pathogenesis additions

rando2:
@ypar had made a branch of possible edits that were never merged in -- I want to compare them to what we have and see if there's anything we can add!

### Checklist


rando2:
Some of this will be redundant with other text we have added since it was written -- but I'd rather have a restructuring problem and keep Yoson's text than have it just hanging out on a branch!

rando2:
I am going to close this but potentially re-open later when we figure out how this maps onto the existing text

PR #592, opened 2020-08-10
Additional comments on mRNA vaccine section

rando2:

I had a few additional comments about possible limitations of this study:
1. For most of the therapeutics, we mention that they were open-label as a possible limitation. Would the same hold here? Obviously it's harder to psychosomatically acquire antibodies...
2. I find it a little worrisome that they didn't test anyone for SARS-CoV-2 or antibodies given the small sample size, although the dosage curve suggests that this wasn't a huge issue. Is this worth mentioning as a possible limitation?
3. In the paper they say: `Though correlates of protection from SARSCoV-2 infection have not yet been determined,
measurement of serum neutralizing activity has been shown to be a mechanistic correlate of protection for other respiratory viruses, such as influenza and respiratory syncytial virus, and is generally accepted as a functional biomarker of the in vivo humoral response.` Would we also want to mention this caveat?

I am happy to add any of the above caveats, but just wanted to ask @cgreene's thoughts before I started adding to the existing text!

### Related issues

#553 

@cgreene 

### Checklist


cgreene:
For the points:
1. This was open label. If the standard is to do that, then I think it's a good idea. It seems implied by dose ranging but I agree it's better to be consistent.
2. Given the timing of enrollment and the Day 1 results and the exclusion criteria, it doesn't look like they probably got folks with a prior infection (at least not one that generated a detectable immune response). For the 30k person study, I imagine they are likely to encounter folks who have already had an infection, so this will probably crop up there (potentially around safety/tolerance). I don't have strong feelings either way.
3. This is going to be a limitation more broadly of early-stage vaccine trials. Do you think it would be better to discuss more broadly in the vaccines section, and then note the specific cases where that is no longer true as the science proceeds?

If you want to piggy-back those changes onto this PR, I'm happy to give it a quick once over! 😁

rando2:
Thanks @cgreene! Lines 778 and 860 were modified to address points 1 & 3.

PR #607, opened 2020-08-15
Update external resources environment

agitter:

This fixes a failure in the scheduled external resources build: https://github.com/greenelab/covid19-review/runs/987098201?check_suite_focus=true

The descartes package wasn't available in the conda environment.  I ran the build scripts locally in the new environment and updated the stats and figures.

### Related issues

#569

### Checklist


ebmdatalab/generate-ebmdatalab-stats.py
```diff
@@ -104,7 +104,7 @@ def assign_ISO(countries):
             else:
                 failed_matches.append(country)
     # Print warning about failures and return successes as dictionary
-    print("Could not assign country codes to: ", ", ".join(failed_matches))
+    print("Could not assign country codes to:", ", ".join(failed_matches))
```
rando2, line 107:
```suggestion
    print("Could not assign country codes to: ", ", ".join(failed_matches))
```
This was your fix that seems to have been reverted!

rando2:
Sorry for the dependencies @agitter!

ebmdatalab/generate-ebmdatalab-stats.py
agitter: I reverted it intentionally because when I ran locally I got two spaces.  It doesn't really matter either way.  It's only a log message in the CI build.

agitter:
No problem @rando2.  My approach with these scheduled builds has been to merge updates and see what happens.  It's hard to anticipate everything that will happen when it runs on GitHub Actions.

ebmdatalab/generate-ebmdatalab-stats.py
rando2: Oh no!!! I'm sorry for changing it back. When I push the fix for the ISO codes, I can put this back the way it was!

PR #611, opened 2020-08-18
Revisions from read-through of full introduction

cgreene:
### Description of the proposed additions or changes

I read through the full introduction, and had some elements that I thought should be revised. In some cases I directly modified them. In other cases, I started a TODO and, in certain cases, a github issue.

### Related issues

I filed #609 and #610 based on this read through.

### Suggested reviewers (optional)

Definitely @rando2! Also others who may have edited these sections and want to contribute.

### Checklist

- [x] Text is formatted so that each sentence is on its own line.

- [x] Pre-prints cited in this pull request have a GitHub issue opened so that they can be reviewed.

PR #613, opened 2020-08-18
Fix broken reference

cgreene:

I noticed a broken reference while reading for structure. This should fix it.

### Checklist


rando2:
EOL characters may cause merge conflict with #611 -- fingers crossed I'm wrong!

cgreene:
I will make sure to get #611 in before this then, since this one is trivial :)

PR #627, opened 2020-08-27
Update EBM script for expanded data frame

agitter:
An upstream change to the EBM data https://github.com/ebmdatalab/covid_trials_tracker-covid/pull/27 added a new "acronym" column.  This causes our EMB updates [to fail](https://github.com/greenelab/covid19-review/runs/1034175398?check_suite_focus=true#step:5:9) because there are more data columns than header entries.

Fortunately, the external resources update is robust enough that the latest correct statistics and figure can still be used in manuscript builds.  However, it might be a good idea to add more automated notifications when parts of that scheduled workflow fail.


rando2:
The run failed but I'm assuming that's because this isn't merged in yet! Pushing ahead and hoping that is the correct interpretation.

agitter:
That's the correct interpretation.  I'll check tomorrow to see if the next scheduled workflow succeeds.

agitter:
The external resources workflow is working again.  15f08ed008b5ebd2cd5b5d6420677175ee590872 updated the EBM stats.

PR #634, opened 2020-09-04
Pathogenesis restructure

rando2:

This is some restructuring for the pathogenesis section. I combined it with the "virology" section and I believe this PR deletes that file.
Not all of the TODOs are addressed, but it seemed better to push this and then edit the scientific content separately.

@cgreene 

### Checklist


content/07.pathogenesis.md
```diff
@@ -61,19 +85,48 @@ Also, MERS-CoV downregulates antigen presentation via MHC class I and MHC class
...
-Understanding the acquisition and retention of antibodies is important to both the diagnosis of prior infections that are no longer active and to the development of vaccines.
-Two of the most common immunoglobulin classes are immunoglobulin M (IgM), which are the first antibodies produced in response to an infection, and immunoglobulin G (IgG), which are the most common antibodies found in the blood.
-Following SARS infection, IgM antibodies were first detected in most patients during the second week following infection and remained detectable until the twelfth week, while IgG antibodies were first detected in the second or third week and remained at close to their peak concentration through the thirteen weeks of the study [@doi:10.1056/NEJM200307313490520].
-A two-year longitudinal study following convalesced SARS patients with a mean age of 29 found that IgG antibodies were detectable in all 56 patients surveyed for at least 16 months, and remained detectable in all but 4 (11.8%) of patients through the full two-year study period [@doi:10.1086/500469].
--->
+#### Cell Susceptibility
```
cgreene, line 88:
```suggestion
#### Host Cell Susceptibility
```
Maybe?

content/07.pathogenesis.md
```diff
@@ -102,82 +155,88 @@ The acute phase response to infection results in the heavy damage of the endothe
...
+They also reported that direct stimulation with inflammatory cytokines such as type I interferons (e.g., IFN&beta;) resulted in the upregulation of ACE2 in human bronchial epithelial cells, with treated groups showing four-fold higher ACE2 expression than control groups at 18 hours post-treatment [@doi:10.1101/2020.02.24.963348].
+G
+Whether SARS-CoV-2 facilitates the positive regulation of its own transmission between host cells is still unclear, the host immune response itself likely plays a key role in mediating infection-associated pathologies.
+A systems-biology approach allows for analyses such as these to identify possible phenotypic and endophenotypic responses to SARS-CoV-2 infection and to develop new hypotheses about how pathogenesis proceeds.
+
+#### Transcriptomics
+
+In addition to the study described above, two other studies have profiled expression following SARS-CoV-2 infection using human cell lines (Figure {@fig:cell-lines-moi}A).
```
cgreene, line 199: This section, including the figure, feels like a lot more detailed dive into a couple datasets than the rest of this review. I leave it to your discretion, but I wonder if it would make the most sense to describe the broad findings here and leave the specifics (actual correlation values, etc) to the source literature.

cgreene (commented):
I had a couple thoughts. I think one is a pretty simple suggestion, but the other is a bigger question. I'm curious what you think on the question.

content/07.pathogenesis.md
rando2: I agree we will probably want to make this section a little less detailed! I had some questions for the original author but figured it would be better to handle that in a separate PR. I don't know what you think about that. The changes here should mostly just be cosmetic because I wasn't quite sure yet how to tie it into the rest of the section.

Also, today someone sent me a [medium post](https://elemental.medium.com/a-supercomputer-analyzed-covid-19-and-an-interesting-new-theory-has-emerged-31cb8eba9d63) (whose representation of several things I have issues with, for the record) but it seems to be addressing some new hype topic (bradykinin) that ties quite closely to some of the material here. 

I think we will want to continue revising this section and figure out exactly what these omics studies say about pathogenesis.

cgreene (approved):
I think this is moving in the right direction - that section I pointed out does feel unusually detailed compared with the rest, but I agree with the plan to revise separately instead of as a big PR.

Issue #610, opened 2020-08-18
Subpopulations of Special Concern Revision


cgreene:
**Is your feature request related to a problem? Please describe.**

I think this section could use a significant revision.

**Describe the solution you'd like**

Here's what I wrote in the TODO I added to the manuscript:
> I think this section could use a revisiting. I think we should deal with the male sex and older age portion in an initial paragraph. After that, it looks like the racial burdens are likely to be attributable in large part to systemic racism - incarceration, wage discrimination, housing discrimination, and more - both now and in the past. Right now this jumps around b/w groups and presents discrimination as the basis in disparities as a possibility but not necessarily a likely contributor. I also worry that it makes many of these things research questions, even though one of the resources we cite ( 10.1093/tbm/ibaa055 ) includes policy recommendations. In short, it feels like we list a bunch of proximal causes but don't tie things back to the overarching likely cause. Finally, the way genetic factors are brought up makes me worry that someone might have, as a takeaway, that genetics may play a driving role in disparities at this level. 

**Describe alternatives you've considered**

We could restructure this in a number of ways. What's here right now feels to me like it wanders around the point more than addressing it.

**Additional context**

dziakj1:
I guess that reorganization would make sense.  I know that @rando2  is working with a student on a related section in the discussion.  When we are discussing research issues, there is another interesting paper we should mention:
Griffith, G., Morris, T. M., Tudball, M. et al. (2020). Collider bias undermines our understanding of COVID-19 disease risk and severity Epidemiology. @doi:10.1101/2020.05.04.20090506

dziakj1:
The equity section in the discussion is discussed in #552 -- I don't know how to determine which information goes in each of the two sections, though.  @tlukan and @rando2 are working on it.

dziakj1:
Also, another important paper just came out on probability of mortality conditional on hospitalization (@doi:doi:10.1001/jamanetworkopen.2020.18039)
with commentary (doi:10.1001/jamanetworkopen.2020.18696)
https://jamanetwork.com/journals/jamanetworkopen/fullarticle/2769387
https://jamanetwork.com/journals/jamanetworkopen/fullarticle/2769381
I haven't read it yet, but apparently the racial disparities appear more in the probability of getting infected and/or hospitalized, not the probability of death conditional on hospitalization.

rando2:
@dziakj1 I have a draft PR in the works and will tag you when I put it up! Thanks for the articles!

PR #638, opened 2020-09-17
Expand on risk due to age, comorbidities, and biological sex

rando2:

I tried to flesh out the text related to age and biological sex ONLY. I am still working on some edits to the other sections in about risk factors (pre-existing conditions, poverty, racism) but it seemed better to do this in pieces so it doesn't get overwhelming.

### Related issues

Partial response to questions raised in #610 

@dziakj1 @tlukan @cgreene 

### Checklist


content/08.transmission.md
```diff
@@ -136,6 +136,42 @@ The sequencing data can be found at GISAID [@https://www.gisaid.org/], NCBI [@ht
...
+Untangling the factors influencing COVID-19 susceptibility is a complex undertaking.
+Among patients who are admitted to the hospital, outcomes have generally been poor, with rates of admission to the intensive care unit (ICU) upwards of 15% in both Wuhan, China and Italy [@doi:10.1056/nejmoa2002032; @doi:10.1001/jama.2020.2648; @doi:10.1001/jama.2020.4031].
+However, hospitalization rates vary by location [@doi:10.15585/mmwr.mm6915e3].
+This variation may be influenced by demographic (e.g., average age in the area), medical (e.g., the prevalence of comorbid conditions such as diabetes), and social (e.g., income or healthcare availability) factors that vary geographically.
+Additionally, some of the same factors may influence an individual's probability of exposure to SARS-CoV-2, their risk of developing a more serious case of COVID-19 that would require hospitalization, and their access to medical support.
+As a result, quantifying or comparing susceptibility among individuals, communities, or other groups requires consideration of a number of complex phenomena that intersect across many disciplines of research.
+
+#### Biological Risk Factors
```
rando2, line 148:
```suggestion
#### Innate Biological Risk Factors
```
What do you call age and sex, especially in relation to comorbidites and "race" (which has both a biological and social component)? I haven't been able to figure this out.

dziakj1: Some people use the word "risk factors" to mean anything that is statistically predictive of an outcome (so being male or Black could be a risk factor).  Sometimes I don't like that because the colloquial meaning of risk implies direct causation, and when people use words like "risk" and "effect" they very often slip into implying that correlation equals causation, even if they are smart.  More recently, some people object to calling race a risk factor because it sounds like a biological explanation and they say that racism is the real cause of the harm.  This objection doesn't make sense if you take risk factor to simply denote a statistical predictor, but it makes a lot of sense if you take risk factor to denote a direct cause.  
Age and sex are also correlated with social and economic experiences, so I don't think that I would say that they are fundamentally different from race in that regard.  
Age, sex, and race can definitely be called important demographic factors.  I wouldn't think of comorbidities as demographic though.  
You could call them all "Patient Characteristics Associated with COVID Risk."
This raises the issue of whether to mention blood type -- I mentioned it as a statistical predictor but someone objected on the grounds that the evidence wasn't consistent enough.  More to the point, it isn't really as useful when making recommendations for policy interventions, because it isn't something that you would ordinarily know about somebody.

dziakj1: I originally called all of these "subpopulations of special concern," but Dr. Greene objected to this -- I think his objection was that age and sex are probably predictive because of biological differences that can't be changed, while race is probably predictive because of social injustices that ought to be intervened on.  I don't really see why this distinction is necessary in the context of this paper, but I respect his judgment.
I guess you could have "Biological Risk Factors" and then "Socioeconomic Risk Factors."  That would make sense.  Of course, that assumes you're willing to use the word "risk factor" at all.

cgreene: As I'm thinking about this now, there are also ways where age and sex might not be strictly innate biological risk factors. In a very constrained health system, triage might lead to different levels of attention being paid to people who were perceived as the least likely to recover. I think it is probably best to treat them all as variables associated with differential outcomes.

I agree that `risk factor` is different if folks take it with a direct causal interpretation. Perhaps a short lead in to the section on risk factors that notes that some may be biological and intrinsic while others are not. Just thinking off the cuff here.

cgreene: That may have been a premature objection. I just wanted to make sure we weren't conflating factors driven by society with those driven by biology.

dziakj1: I guess that if we want to be really pedantic, we shouldn't use the word "innate" for age, because age can change, and innate is Latin for "born that way."  But "intrinsic" might be okay.

dziakj1: I do think it might be better to think of sex as a purely biological predictor here though -- It's true that women are more likely to do caregiving work in our society, but that wouldn't explain why *men* were at higher risk of death.  As @rando2  mentioned, age definitely does have a social dimension because if you're in a nursing home and there's an outbreak, you're in real trouble.

SiminaB: In general we know that women do better with many illnesses compared to men and have higher life expectancy (seemingly across the world) and that may be related to immune system differences (due to women's immune systems being able to carry pregnancies for example; women are also more likely to get more auto-immune diseases, but the immune benefits definitely seem to outweigh this.)

rando2: Thank you all so much for these thoughts! I really appreciated the break down of the causal implication that people may take, and I agree proactively addressing it seems wise! I tried to address the feedback in the latest commit by adding a few lines to the intro paragraph and changing the title to "Intrinsic Traits Associated with Increased Risk" (?)

dziakj1: If you want, you could say "Patient Traits" instead of "Intrinsic Traits" so that all the philosophers eager to argue about exactly what is and is not intrinsic can take the day off -- It's up to you though.

rando2: I love it @dziakj1! I definitely do not want to argue with philosophers!

rando2:
Thank you so much everyone for the feedback! @cgreene or @dziakj1 would you be able to approve if you think this is OK to merge?

dziakj1 (approved):
It looks very good, @rando2 !

rando2:
Thank you so much @dziakj1! It's a lot easier to build off of what's already there than to start from scratch!

PR #639, opened 2020-09-23
Address *correlation* between ancestry and COVID-19 risk

rando2:

This PR expands on some of @dziakj1's text about race and ethnicity's association with negative outcomes. @cgreene had raised the concern that we should make sure we're super clear that we separate out the genetic and social correlates of ancestry (namely, it's very unlikely genetics are playing a big part, and VERY likely that systemic racism is). 

I tried to blend the biological anthropology in, but I'm actually not totally sure what most mSystems readers will know about human history -- please let me know if it that part is confusing. Basically the groups who keep showing up at risk are located on totally different parts of the human family tree, and it would be quite surprising if indigenous American groups shared genetic risk factors with "Africans" that weren't found in East Asians. (Especially since you can't treat Africa as a single group when talking about genetics because there is so much diversity).

Some of the social factors will be coming tomorrow -- almost there but want to look at it again with fresh eyes tomorrow.

### Related issues

#610 

@cgreene @dziakj1 @tlukan @juliettemarie0405 @ipsitak @shiktadas

### Checklist


content/08.transmission.md
```diff
@@ -136,57 +136,27 @@ The sequencing data can be found at GISAID [@https://www.gisaid.org/], NCBI [@ht
...
+However, such variants are unlikely to explain the patterns that are appearing on a global scale.
+Reports from several sub-Saharan African countries have indicated that the effects of the COVID-19 pandemic have been less severe than expected based on the outbreaks in China and Italy.
+In Kenya, for example, estimates of national prevalence based on testing blood donors for SARS-CoV-2 antibodies were consistent with 5% of Kenyan adults having recovered from COVID-19 [@doi:10.1101/2020.07.27.20162693].
+This high seroprevalence of antibodies lies in sharp contrast to the low number of COVID-19 fatalities in Kenya, which at the time was 71 out of 2093 known cases [@doi:10.1101/2020.07.27.20162693].
+Likewise, a serosurvey of health care workers in Blantyre City, Malawi reported an adjusted antibody prevalence of 12.3%, suggesting that the virus had been circulating more widely than thought and that the death rate was up eight times lower than models had predicted [@doi:10.1101/2020.07.30.20164970].
```
rando2, line 156: My issue with this preprint is that they assume that 12.3% rate measured in healthcare workers extrapolates to the whole population. So their 8x estimate they give is likely a bit too rosy. I don't know if the phrasing "up to" does a sufficient job of providing a bit of skepticism about this number without undermining their interesting findings.

SiminaB:
My understanding is that for individuals hospitalized with COVID19 the risk of dying is similar for Black and non-Black individuals in the US and that a higher proportion of Blacks vs. non-Blacks are infected (I've seen presentations on this - will look for references in a bit). In other words, the disparity is more at the level of infection, so even if there were a genetic variant at play (which I agree, is very unlikely), it would somehow impact the risk of being infected. I would assume that the risk of being infected being higher in Blacks is mostly due to the types of jobs they have (more essential workers, for instance) as well as their housing conditions (more in public housing and multigenerational homes, fewer in single-family homes.)

We know that there are also a number of comorbidities that strongly influence risk of death, so I would think those are also more common in Blacks vs. non-Blacks, as most health disparities are, so I'm not sure if that also taken into consideration when looking at risk of death. It's possible that once someone is in the ICU, for example, diabetes and other comorbidities no longer matter as much.

SiminaB:
Here is a reference https://jamanetwork.com/journals/jamanetworkopen/fullarticle/2769387/ Will add it in a bit as an issue. They do a bunch of adjustments, but even w/o them, the hospital mortality is actually higher in whites compared to Blacks. However, whites are more likely to be older and male, which are 2 major factors for poor outcomes, so the fact that they cancel out may point to structural racism being a factor for Blacks. The fact that most of the Blacks are women may be related to the fact that they tend to have more essential jobs.

shiktadas:
@SiminaB yes genetically it will be impossible to say that the ACE2 are different between ethnic groups. (I am not sure this kind of comparison is done yet). SARS-CoV-2 directly infect endothelial cells by binding to their ACE2 receptors. The common theme in individuals with co-morbidites is the vascular inflammation nad endothelial dysfunctioning. These are higher in ethnic groups in comparison to whites due to hypertension, diabetes, obesity etc. And why do they have higher co-morbidites because of structural health inequalities = racism. All I am trying to say is that this virus is exposing both the physiological differences to infectiousness as well as social causes for it.

shiktadas:
The mortality is higher in whites because they are tested more. Also they live in affluent areas where there are facilities to test and go to hospital. The death will not be recorded as COVID-19 related if you died at home.

SiminaB:
@shiktadas I agree with your points in general, including racism = structural inequalities. However, in the paper I linked they look at mortality for individuals who are hospitalized, so it doesn't include people who die at home. Being tested more could also imply lower mortality (CFR) if many more asymptomatic/mild cases are found compared to the number of cases who died from COVID19. I've seen this type of result in several other presentations from various hospital systems - the inequality manifests itself in being hospitalized with severe disease, not so much once one is hospitalized, which makes sense because Blacks have more opportunities to get infected + more comorbidities that make them more likely to have severe disease, hence to be hospitalized.

shiktadas:
From UK we have good evidence : I haven't followed outside UK data - https://www.cebm.net/covid-19/bame-covid-19-deaths-what-do-we-know-rapid-data-evidence-review/

rando2:
@SiminaB @shiktadas thank you both so much! I made some adjustments to the text to incorporate the sources you linked. @SiminaB that is especially fascinating the extent to which SES can explain it alone, I have been putting together a section about the effects of chronic stress on the risk of developing comorbidities but I'll have to make sure we really need it!

I also added a little bit more about the possibility that there *could* be some genetic variants after talking with a biological anthropologist today.

content/08.transmission.md
```diff
@@ -142,21 +142,31 @@ The sequencing data can be found at GISAID [@https://www.gisaid.org/], NCBI [@ht
 
 A number of studies have suggested associations between individuals' racial and ethnic backgrounds and their COVID-19 risk.
 In particular, Black Americans are consistently identified as carrying a higher burden of COVID-19 than white Americans [@doi:10.1377/hlthaff.2020.00598; @doi:10.15585/mmwr.mm6925e1], with differences in the rates of kidney complications from COVID-19 particularly pronounced [@doi:10.1016/j.kint.2020.05.006].
+Statistics from a number of cities indicate significant discrepancies between the proportion of COVID-19 cases and especially deaths in Black Americans relative to their representation in the general population [@doi:10/ggv494].
```
cgreene, line 194: What is especially highlighting here? I think it could just be _cases and deaths_...

cgreene (approved):
I had one quick comment/question, but this looks like an improvement so merge when you're ready 👍

rando2:
Planning to merge this in, but additional PRs to improve the text are of course welcome!

Issue #662, opened 2020-11-03
Please review the manuscript!


rando2:
Hi everyone,
The manuscript is at a point where it seems like we can move ahead with submitting version 1 to _mSystems_! Here's what we need from you:
1. Please look through/review [the manuscript](https://greenelab.github.io/covid19-review/) to ensure that you feel comfortable attaching your name to it. Please try to do this -- or at least let us know if you can't -- by EOD on Wednesday, November 11th. We would love to try to submit by the end of next week!
2. Open PRs to address any issues you think need to be changed before submission. (I am at a conference this week but will do everything I can to review/approve PRs as quickly as possible! Anyone who is willing to review PRs coming in this week will be greatly appreciated!)
3. Check that your [author information](https://github.com/greenelab/covid19-review/blob/master/content/metadata.yaml) is correct, especially:
- Conflict of Interest
- Roles
- Code of Conduct Approval
4. Open a PR updating the date of your COI approval to [date that you're doing this] so that we know you are ready for submission!

Thank you so much everyone, I am really excited to be moving forward to the next step! Please feel free to ask questions below.
Best wishes,
Halie

nafisajadavji:
@rando2 How do we approve the Code of Conduct?

rando2:
@nafisajadavji Here you go!
1. Go to the [metadata.yaml file](https://github.com/greenelab/covid19-review/edit/master/content/metadata.yaml)
2. Click on the pencil in the upper right corner of the page to edit
3. Look for your name
4. Edit the No on line 365 to say Yes
<img width="773" alt="Screen Shot 2020-11-04 at 2 05 50 PM" src="https://user-images.githubusercontent.com/8975841/98157124-db71ba00-1ea6-11eb-8c0f-ea7d78dd243d.png">
Let me know if you need any help!

nafisajadavji:
@rando2 Thanks! I think I did it correctly. I am still learning :S

dziakj1:
Thanks @rando2 !  My information is up to date and I would still like to be included as a co-author if that is okay.  I'm sorry I haven't been much help since the summer.

rando2:
Hi @dziakj1, thank you so much! You absolutely qualify for authorship if you're interested! If you are able to adjust the date of your Conflict of Interest approval, that is the best way for us to track who is happy to move forward with being an author -- if you need help, I can make a PR and tag you for review!

PR #663, opened 2020-11-03
Add code of conduct field to metadata for all authors

rando2:

Because we added the code of conduct late, not all authors have this field in their metadata.
This PR just adds it with "no" as the default value.

### Checklist


cgreene (approved):

cgreene:
I agree with this, and I went ahead and merged to get ahead of PRs from #662

PR #668, opened 2020-11-03
A new paper about vitamin D deficiency from last week

rando2:

Hey Ronan, I saw this paper and thought it seemed interesting and might worth adding as a citation if you think it fits!

### Checklist


RLordan (approved):
Perfect @rando2 , thanks for updating I only just read it myself. There is now 17 observational studies looking at the association between vit D and COVID. only 2 were skeptical of the association. The evidence is building.

rando2:
It's super interesting for sure! I kind of wonder if there could be some conditions associated with vitamin D deficiency and also with increased susceptibility to COVID-19, but it definitely seems worth it to control what we can control (aka vitamin D status!)

RLordan:
> It's super interesting for sure! I kind of wonder if there could be some conditions associated with vitamin D deficiency and also with increased susceptibility to COVID-19, but it definitely seems worth it to control what we can control (aka vitamin D status!)

You are completely correct, risk factors like CVD, cancer, diabetes, hypertension, and age all associated with low vit D, but in the former conditions rcts not so confident that supplementation helped... particularly for CVD, very mixed results.

rando2:
@RLordan that's super interesting!!! You are always welcome to press merge in these situations, I just check periodically to see if the builds worked (all green at the bottom) so I'll hit it now :)

PR #681, opened 2020-11-10
Minor text edits

ajlee21:

This PR is making very minor text edits after reading through the manuscript as a whole.

Congrats on this manuscript coming together! It was great to read and learn so much about covid. Having the text organized into these different sections really helped. 

A couple of comments:
1. I noticed a bunch of missing citations in section 3, but I wasn't sure how to fix them..they looked right to me just glancing through...
2. In section 3, it doesn't intuitively make sense to me that you can start being infections at day 12 before symptom onset when your mean incubation period is 5 days. I guess My understanding is that mean incubation period of 5 days means that  you acquired the infection and 5 days later you started showing symptoms. So 12 days prior to symptom onset is before you acquried the infection. I'm sure I'm misunderstanding something here but I wanted to check.

### Checklist


rando2 (approved):

rando2:
Good catches Alex! And thank you for the tip about the citations. I'm guessing these overlap with #682 so linking that here! 

For the incubation period/contagiousness, I think the issue is that there are three stages:
A. Someone has been exposed to SARS-CoV-2 but the infection has not progressed enough to be detected by PCR or manifest as symptoms
B. Infection has progressed enough to be detected by PCR (end of incubation period)
C. Infection has progressed enough to present with symptoms (symptom onset)

The relationship between B & C is super fuzzy. It seems like some people get symptoms before their PCR test comes back positive, and it seems like others get a positive before getting symptoms. The incubation period is (if I remember correctly) 5-9 days on average, so I think this is within the realm of possibility that some people could be positive 12 days before symptoms onset, depending on the relationship between viral load and contagiousness, which we also don't know. 

The research on this topic definitely all has different numbers, which makes me think a lot more precise measurements are needed. I think that 12 days one might have been quite different than the other estimates? I can take another look at this section but figured I'd share my initial thoughts in case they're helpful!

ajlee21:
Thank you for the thorough explanation @rando2 ...I don't think this needs to hold up submission, but something that I found curious. I guess I always assumed there was just 2 stages (A and C) so I lumped B and C together. I didn't realize that there could be an additional delay between viral detection and symptoms (at least if you're symptomatic). I would definitely need to read more on this, but something I would agree is probably hard to get precise measurements for

PR #683, opened 2020-11-10
Update rootstock on 2020-11-10

agitter:
This rootstock update patches two severe issues in the build environment.  One is causing builds to fail now.  The other may cause builds to start failing in a week.


rando2 (approved):
Wonderful, thank you so much @agitter! I will try to keep an eye out to merge this one first, before anything else gets merged in.

Issue #682, opened 2020-11-10
Broken citations


agitter:
The [build log](https://github.com/greenelab/covid19-review/runs/1374429899) for 24da920c66a20b3970440174070a8beee0e5301e shows warnings and errors for the following citations:
- doi:10.1093/pcmedi/pbaa032/5901532
- doi:10.2307/26061262
- doi:10.23812/20-editorial_1-5
- doi:10.2807/1560-7917
- doi:10.4415/ann_13_02_03
- https://act.nationalnursesunited.org/page
- https://mronline.org/wp-content/uploads/2020/06/2020-04-Frontline-Workers.pdf
- https://www.cdc.gov/about/advisory/pdf/VentDocument_Release.pdf
- https://www.census.gov/content/dam/Census/library/publications/2019/demo/p60-267.pdf
- https://www.issuelab.org/resources/695/695.pdf
- https://www.physicianscientists.org/page/summer-research-pilot-program
- https://www.synairgen.com/wp-content/uploads/2020/03/200318-Synairgen-to-start-trial-of-SNG001-in-COVID-19-imminently
- https://www.synairgen.com/wp-content/uploads/2020/07/200720-Synairgen-announces-positive-results-from-trial-of-SNG001-in-hospitalised-COVID-19-patients.pdf
- https://www.who.int/publications/i/item/report-of-the-who-china-joint-mission-on-coronavirus-disease-2019

Searching the HTML manuscript for "???" can help us find many of them.

agitter:
doi:10.2807/1560-7917 appears to be a partial DOI from Eurosurveillance.  It was added in #635.  @rando2 do you know which paper it refers to?

doi:10.2307/26061262 is not a valid DOI, but I'm not sure where it was originally added.  It was moved out of 08.transmission.md in #659.

I'm less concerned about the URL references.  They don't have nice metadata in the reference list, and a couple have nonsense, but they provide the correct URL:
![image](https://user-images.githubusercontent.com/3532898/98725987-7d881b00-235b-11eb-9555-c8b643b1c212.png)

rando2:
Thank you so much @agitter! Both of those were me. The second one (10.2307/26061262) was a review article by Robert Sapolsky about chronic stress & poverty -- I don't 100% remember which one I was trying to cite because he has written a lot on this topic, but I was aiming just to give people a high-level, general primer to check out if they want to learn more about a well-studied topic, so the reference that I have swapped it for should work regardless of whether it was the original intent!

[...]

PR #687, opened 2020-11-10
Correct broken references

agitter:

This fixes some of the broken DOI references.  `doi:10.4415/ANN_13_02_03` is a valid DOI, but data cite doesn't have the metadata.  I switched to PMID.  I commented in #682 with the DOIs I couldn't resolved.  We can add those here before merging.

### Related issues

#682

@rando2 please double-check that I'm citing the correct papers.  

### Checklist


rando2 (approved):
Thank you for fixing these @agitter! I checked and all the changes pull up the correct articles for me.

rando2 (approved):
There is one more ?? that I'm seeing in the document, but since it's figure-related, I'll add that to the figure PR!

PR #725, opened 2020-11-16
Fix typo in 20.treatments.md

lubianat:
### Checklist


agitter (approved):
Good catch!

PR #728, opened 2020-11-17
Update viral nomenclature per journal guidelines

rando2:

The journal guidelines are as follows:

> **Viruses.** Names used for viruses should be those approved by the International Committee on Taxonomy of Viruses (ICTV) and reported on the ICTV website. In addition, the recommendations of the ICTV regarding the use of species names should generally be followed: when the entire species is discussed as a taxonomic entity, the species name, as with other taxa, is italic and has the first letter and any proper nouns capitalized (e.g., _Tobacco mosaic virus_, _Murray Valley encephalitis virus_). When the behavior or manipulation of individual viruses is discussed, the vernacular (e.g., tobacco  mosaic virus,  Murray Valley  encephalitis virus) should be used. If desired, synonyms may be added parenthetically  when  the  name  is  first  mentioned.  Approved  generic  (or group) and family names may also be used.

Additionally, @lubianat noted that we should use SARS-CoV-1 instead of SARS-CoV, so I have made those modifications here as well!

### Checklist


agitter (approved):
Looks good to me.  Let's wait for the spellcheck report before merging.

PR #740, opened 2020-12-17
Add updated Pfizer/BioNTech info based on recent paper/press releases

rando2:

This is an update to the Pfizer vaccine, which hasn't been changing as much the past couple weeks. Additional updates very very welcome. Possible topics include: safety, progress in rollout (which countries, for example), accessibility (cost, groups who are excluded)

### Related issues

#737 

### Checklist


content/20.pharmaceuticals.md
```diff
@@ -712,6 +712,24 @@ Participants reported mild and moderate systemic adverse events after the day 1
...
+A preliminary report was published in the _New England Journal of Medicine_ one month later [@doi:10.1056/NEJMoa2034577].
+The vaccine candidate is comprised of the full SARS-CoV-2 spike protein in a vaccine formulation based on modRNA technology [@doi:10.1101/cshperspect.a014035].
+During the Phase III trial, 43,538 participants were enrolled 1:1 in the placebo and the vaccine candidate and received two 30-μg doses 21 days apart [@doi:10.1056/NEJMoa2034577].
+Of these enrolled participants, 21,720 received BNT162b2 and 21,728 received a placebo [@doi:10.1056/NEJMoa2034577].
+Recruitment occurred at 135 sites across six countries: Argentina, Brazil, Germany, South Africa, Turkey, and the United States.
+An initial press release described the first 94 cases, which were consistent with 90% efficacy of the vaccine at 7 days following the second dose [@url:https://www.pfizer.com/news/press-release/press-release-detail/pfizer-and-biontech-announce-vaccine-candidate-against].
```
ajlee21, line 722: So looks like efficacy is defined based on no onset of symptoms at day 7 for those without covid19 initially, or subsiding of symptoms for those with covid19 initially. I'm not a pharmacology expert so I was just wonder if this is a standard definition. I'm curious why they don't use viral load instead since patients can be asymptomatic but still shed.

ajlee21: I guess if this isn't the standard definition, I might add text to describe what is meant by efficacy in this case.

content/20.pharmaceuticals.md
```diff
@@ -712,6 +712,24 @@ Participants reported mild and moderate systemic adverse events after the day 1
...
+The release of the full trial information included additional cases and analyzed the first 170 cases occurring at least 7 days after the second dose, 8 of which occurred in patients who had received BNT162b2.
+The first interim analysis reported the results of the first 94 cases of COVID-19 occurring at least 7 days after the second dose.
+Of these, 8 occurred in participants who received BNT162b2 and 162 occurred
```
ajlee21, line 725: Is this sentence unfinished?

content/20.pharmaceuticals.md
```diff
@@ -712,6 +712,24 @@ Participants reported mild and moderate systemic adverse events after the day 1
...
+The primary efficacy analysis of the Phase III study was concluded on November 18, 2020 [@url:https://www.pfizer.com/news/press-release/press-release-detail/pfizer-and-biontech-conclude-phase-3-study-covid-19-vaccine], and the final results indicted 94% effectiveness of the vaccine.
+The vaccine was approved in early December by the United Kingdom's Medicines and Healthcare Products Regulatory Agency with administration outside of a clinical trial beginning on December 8, 2020 [@doi:10.1136/bmj.m4714; @url:https://www.bbc.com/news/uk-55227325].
+As of December 11, 2020, the United States FDA approved this vaccine under an emergency use authorization [@url:https://www.fda.gov/news-events/press-announcements/fda-takes-key-action-fight-against-covid-19-issuing-emergency-use-authorization-first-covid-19].
+ 
 In summary, mRNA vaccines are promising tools in the prevention and control of pandemics because they may sidestep many challenges associated with traditional vaccine design and manufacturing.
```
ajlee21, line 733: Just curious if you know after reading about these vaccines, why it is that these mRNA vaccines are more effective than say the flu vaccine? My understanding is that they are both inducing an immune response to prime your body, but the flu is using an attenuated/dead version of the vaccine while these mRNA vaccines are having your cells produce just the antigen (single protein, not even the full viral particle). So the end result is the same but the mechanism of delivery is different - at least that is my take

content/20.pharmaceuticals.md
```diff
@@ -743,6 +743,24 @@ Participants reported mild and moderate systemic adverse events after the day 1
...
+During the Phase III trial, 43,538 participants were enrolled 1:1 in the placebo and the vaccine candidate and received two 30-μg doses 21 days apart [@doi:10.1056/NEJMoa2034577].
+Of these enrolled participants, 21,720 received BNT162b2 and 21,728 received a placebo [@doi:10.1056/NEJMoa2034577].
+Recruitment occurred at 135 sites across six countries: Argentina, Brazil, Germany, South Africa, Turkey, and the United States.
+An initial press release described the first 94 cases, which were consistent with 90% efficacy of the vaccine at 7 days following the second dose [@url:https://www.pfizer.com/news/press-release/press-release-detail/pfizer-and-biontech-announce-vaccine-candidate-against].
+The release of the full trial information included additional cases and analyzed the first 170 cases occurring at least 7 days after the second dose, 8 of which occurred in patients who had received BNT162b2.
+The first interim analysis reported the results of the first 94 cases of COVID-19 occurring at least 7 days after the second dose.
+Of these, 8 occurred in participants who received BNT162b2 and 162 occurred
```
rando2, line 756:
```suggestion
```

content/20.pharmaceuticals.md
rando2: Thanks for catching this! I think I failed to delete the original two lines from before the new data was released.

content/20.pharmaceuticals.md
rando2: That is also my understanding about the mechanism! I'm really not sure why they work so well, especially since I believe people were hoping for much lower efficacy (like 70%) based on just the mechanism, prior to testing! I can look into this but also tagging @RLordan in case he has any ideas!

content/20.pharmaceuticals.md
rando2: This is a great question and I have NO idea. I can look into this!

content/20.pharmaceuticals.md
ajlee21: Sorry don't mean to give you more work. This was more a question that I was curious about in case you knew off hand. But not to say you should look into this, because I can certainly do that on my own:)

content/20.pharmaceuticals.md
rando2: Vaccine efficacy has a [standard definition](https://www.cdc.gov/csels/dsepd/ss1978/lesson3/section6.html), where you are comparing the in your trial and control groups.

In the case of Pfizer, they [registered this timeframe in their defined primary outcomes](https://clinicaltrials.gov/ct2/show/NCT04368728) (e.g., `Confirmed COVID-19 in Phase 2/3 participants without evidence of infection before vaccination [ Time Frame: From 7 days after the second dose of study intervention to the end of the study, up to 2 years ]`)

content/20.pharmaceuticals.md
rando2: I'll keep an eye out and let you know if I see anything!

RLordan: Sorry @rando2 @ajlee21 , I am late to the convo. You are correct with the mechanisms. In both cases you are showing the immune system what the target virus is using different methods. However, I think the greater efficacy comes from how the mRNA vaccine tends to lead to a greater response because the spike protein is made by the cells causing higher doses than a traditional adenovirus vector vaccine could achieve. That’s my take on it, I don’t think they were expecting such high efficacy, which means there will likely be further study into how long these vaccines can maintain their effects etc.

RLordan (approved):

PR #760, opened 2021-01-14
Rough outline of evolution manuscript

rando2:

This is a rough outline of the sections we will want to add to get a standalone evolution manuscript. Early on there were some discussions of zoonotic transfer, etc. in the issues that we might want to make sure we read while developing this!

@Sergey-Knyazev

### Checklist


RLordan:
Do you want to include a bit about the B.1.1.7 variant and other variant of concern here? I can start part of it this week.

Sergey-Knyazev (approved):

RLordan:
Very cool figures of the B.1.1.7 variant (and others) detection that might be worthwhile incorporating

https://cov-lineages.org/global_report_B.1.1.7.html

rando2:
@rlordan that would be fantastic! @sergey-knyazev and I had briefly discussed the UK and SA variants on #750 so I think we'd both be excited to give feedback!

RLordan:
> @RLordan that would be fantastic! @Sergey-Knyazev and I had briefly discussed the UK and SA variants on #750 so I think we'd both be excited to give feedback!

Perfect, i will get on it!

PR #764, opened 2021-01-15
Update @cgreene and @rando2 affiliation

rando2:
@cgreene I am assuming we should keep the Penn affiliations and just add the CU one? Also, I took a stab at adding your affiliation at the Center for Health AI, but the info might not be correct!

### Checklist


cgreene (approved):
Got it!

PR #786, opened 2021-01-19
Added text about spike protein evolution

Sergey-Knyazev:
### Checklist


rando2:
@Sergey-Knyazev this looks good to me! I just wanted to clarify because that paper made it sound like N501Y was one of two mutations rather than the name of the variant itself (and that the N501Y mutation had been around since April or so). I also added a couple of citations about how it has spread to lots of different countries. If these look like OK changes, I'm happy to merge!

PR #763, opened 2021-01-14
Restructure build for individual pathogenesis manuscript

agitter:
### Description of the proposed additions or changes

This pull request will experiment with build script modifications.  The goal is to build a DOCX of only the pathogenesis content.

### Related issues

#732 #762

### Suggested reviewers (optional)

@rando2

### Checklist

- [x] Text is formatted so that each sentence is on its own line.

- [x] Pre-prints cited in this pull request have a GitHub issue opened so that they can be reviewed.

Issue #762, opened 2021-01-14
Build stand-alone pathogenesis manuscript 


rando2:
**Is your feature request related to a problem? Please describe.**
Because we have split everything up, we will need to build docx versions of the manuscripts for submission. We're almost at that point with pathogenesis (#744).

**Describe the solution you'd like**
This will require a few pieces:
1. Select info from metadata.yaml depending on the unique author list for this manuscript
2. Generate a box defining all authors who belong to the consortium (how exactly do we want to phrase this?)
3. Render text from 07.pathogenesis.md only
4. Generate reference list from 07.pathogenesis.md text (see #732)

**Describe alternatives you've considered**
Some manual hijinks with docx involving a lot of copy and paste -- but that would not really work for the reference list.
Could also copy the text into a separate manubot project

**Additional context**

agitter:
Because pathogenesis is being submitted first, I suggest we treat it as a special case.  If we get it to work, we can generalize to generate all of the individual manuscripts.  I'll make a pull request to get started.

We can add additional keys to each author in the metadata to help us select the named authors for each manuscript.  Then the front matter for each manuscript can select only those authors.  Imagine something like
```
authors:
  -
    name: Author A
    contributions:
      - Project Administration
      - Writing - Original Draft
      - Writing - Review & Editing
      - Methodology
    manuscripts:
      - pathogenesis
      - evolution
   -
    name: Author B
    contributions:
      - Writing - Original Draft
      - Writing - Review & Editing
    manuscripts:
      - pathogenesis
      - therapeutics
```

I'm leaving out most of the existing fields.  That assumes the contributions are the same for all manuscripts, which is not really true.  I'm also not yet sure how to order authors.  Maybe we need to manually list the order for each manuscript
```
authors:
  -
    name: Author A
    contributions:
      - Project Administration
      - Writing - Original Draft
      - Writing - Review & Editing
      - Methodology
    manuscripts:
      pathogenesis: 1
      evolution: 5
   -
    name: Author B
    contributions:
      - Writing - Original Draft
      - Writing - Review & Editing
    manuscripts:
      Pathogenesis: 3
      Therapeutics: 6
```

If we need to support manuscript-specific contributions, we could nest those too:
```
authors:
  -
    name: Author A
    pathogenesis:
      order: 1
      contributions:
        - Project Administration
        - Writing - Original Draft
        - Writing - Review & Editing
        - Methodology
    evolution:
      order: 5
      contributions:
        - Project Administration
[...]

PR #793, opened 2021-01-21
 Build multiple individual manuscripts

agitter:

This adds support for building multiple individual manuscripts instead of only the pathogenesis manuscript.  It will build each manuscript listed in `content/individual-manuscripts.txt`.  That file had one keyword per line, where the keyword corresponds to a content markdown file.  I added the evolution manuscript as an example and to test everything works on GitHub Actions.  We can delete that line before merging.

I also updated the methods to mention this build strategy and remove figure references that are broken in the individual methods manuscript.

### Related issues

Follows up on #763 and addresses a TODO from #762

@rando2 

### Checklist


rando2 (approved):
This all looks good to me, although I wasn't able to figure out how to see the output branch off a non-master branch! I also suggested an update to the output/README.md file

agitter:
Good catch on updating the output readme.  Your text looks good to me.

We can't directly view the output branch, but the contents are all zipped and stored as the artifact of the GitHub Actions build.  The latest one is https://github.com/greenelab/covid19-review/suites/1875081278/artifacts/36558292.

Before merging, I'll remove the evolution manuscript from the list to build because it doesn't have the real author list.

rando2:
Thank you so much @agitter! That makes total sense!

PR #790, opened 2021-01-20
Revisions for Pathogenesis from @alavendelm

rando2:
### Description of the proposed additions or changes

@alavendelm sent these suggestions by email -- I made in-line changes as applicable on the command line, and will add in the comments ASAP!

### Checklist

- [ ] Text is formatted so that each sentence is on its own line.

- [ ] Pre-prints cited in this pull request have a GitHub issue opened so that they can be reviewed.

PR #797, opened 2021-01-21
Add author order & contributions for Pathogenesis ms

rando2:

This PR attempts to translate the list of author contributions in #744 into metadata fields.
I bumped people up based on the last week of contributions, so it is not exactly what is in the PR.
@cgreene and @agitter are currently listed as co-last for this paper, but I'm happy to adjust depending on what you want!

### Checklist


rando2:
I am so confused why 8ea0d79 is showing up here given that it's not in my local git log and the changes aren't actually appearing in files. I guess we should wait until #790 merges just to be safe...

cgreene (approved):
🚀

agitter (commented):
I don't need to be co-last, and I definitely shouldn't be first.  I think @rando2 should be first.  If we move @cgreene to `order: 9` I could be `order: 8`, but I'm open to anything.

I'll need to have contributions copied into the `pathogenesis` manuscript metadata.  Do these seem appropriate?
```
contributions:
      - Methodology
      - Software
      - Project Administration
      - Writing - Review & Editing
```

Where should the "COVID-19 Review Consortium" author be positioned?

Do we want to add a footnote about how authors with similar contributions were ordered alphabetically?

Edit: I saw that this was merged.  I can open a new pull request once we agree on the changes.

rando2:
@agitter those definitely sound like appropriate contributions to me!

And yes, that footnote is a great idea!

PR #799, opened 2021-01-22
Fix terminology about ACE2 "receptor"

rando2:

@byrdjb pointed out the term ACE2/hACE2 "receptor" isn't quite right! This just rephrases those to remove the implication that ACE2 is usually acting as a receptor.

### Checklist


cgreene (approved):
Looks good! 🚀

PR #800, opened 2021-01-22
Update pathogenesis metadata and author order footnote

agitter:

Implements the changes from https://github.com/greenelab/covid19-review/pull/797#pullrequestreview-574408995  I wasn't sure where to stick the consortium author.  Is there a better place?

### Related issues

#797

@rando2 

### Checklist


rando2 (approved):

rando2:
The content here looks good but it seems like we might want to check `build/update-author-metadata.py` line 72
`individual_contributions = author["manuscripts"][manuscript].get("contributions", MISSING_CONTRIBUTIONS)`
AttributeError: 'list' object has no attribute 'get'

agitter:
Good catch.  I messed up the contribution indenting.  I fixed that an added a more interpretable error message in case that happens again.

rando2:
Oh, that makes so much more sense! I was a bit confused how we could possibly be getting a different data structure than on previous builds 😂

cgreene:
I feel like I'm used to the consortium author after all of the folks who played major roles in writing but before those who were more on the organizing side (i think the idea being that the consortium author was the first of the last authors)... but that may just be the subfield that I have been part of.

rando2:
@cgreene I think that's consistent with where Tony put it -- 7 is mostly people who edited or provided some feedback, but the contributors in the "last authors" category would be you and Tony.

agitter:
I pushed e4a89b2 to fix a whitespace issue.  This should now be ready to merge.

PR #809, opened 2021-01-25
Superficial fixes before submission

rando2:

I did one final sweep to fix things like abbreviations used < 3 times, end of line whitespace, and sentences missing words.
Also tried to minimize implications that ACE2 is a receptor, since @byrdjb has pointed out that it is not!

### Checklist


content/07.pathogenesis.md
```diff
@@ -274,37 +274,35 @@ For example, because of the high level of sequence homology between SARS-CoV-2 a
...
 
-Considering the paucity of omics-level big data sets for SARS-CoV-2 currently available, existing data hubs that contain information for other coronaviruses such as UniProt, NCBI Genome Database, The Immune Epitope Database and Analysis Resource (IEDB), and The Virus Pathogen Resource (ViPR) will serve as useful resources for comparative bioinformatics research of SARS-CoV-2.
-Using such databases, the systems-level reconstruction of protein-protein interaction (PPI) networks will enable the generation of hypotheses about the mechanism of action of SARS-CoV-2 and suggest potential drug targets.
+Considering the paucity of omics-level big data sets for SARS-CoV-2 currently available, existing data hubs that contain information for other coronaviruses such as UniProt [@doi:10.1093/nar/gky1049], NCBI Genome Database [@url:https://www.ncbi.nlm.nih.gov/genome/], The Immune Epitope Database and Analysis Resource [@doi:10.1093/nar/gky1006], and The Virus Pathogen Resource [@doi:10.1093/nar/gkr859] will serve as useful resources for comparative bioinformatics research of SARS-CoV-2.
```
rando2, line 277: This was suggested in a word-doc feedback I got and I thought I did it, but I guess i somehow did not propose the suggestion because it doesn't show up on any of the PRs!

ajlee21, line 277: Totally fine if you want to remove abbrev here. I'll just say there I know for virologists..these abbrev are very well known and common..though i guess this manuscript will be fore a general audience so that makes sense

ajlee21 (approved):
Good catches!

content/07.pathogenesis.md
rando2: I am not sure how to handle this because I agree they are (and like should I change NCBI to National Center for Biotechnology Information? seems silly!) I guess I was hesitant because the mSystems guidelines specifically say don't give an acronym unless you use it 3 times!

cgreene: They'll find this when they do proofs 🤷  - i think we ship it :shipit:

rando2:
Passed manubot prior to updating for #808, which did not edit any of the same files

PR #741, opened 2020-12-17
Update Moderna vaccine info

rando2:
### Description of the proposed additions or changes

The Moderna vaccine is the trickiest one because they seem to still be releasing new info as they petition for an EUA. Anyone else wanting to contribute to this is extremely welcome! I only got through a little of the recently released data (<https://www.fda.gov/advisory-committees/advisory-committee-calendar/vaccines-and-related-biological-products-advisory-committee-december-17-2020-meeting-announcement#event-materials>). If they release a new paper at some point, this would probably be the easiest way to get this info!

### Checklist

- [ ] Text is formatted so that each sentence is on its own line.

- [ ] Pre-prints cited in this pull request have a GitHub issue opened so that they can be reviewed.

PR #828, opened 2021-02-02
Vaccines outline

rando2:

This is a suggestion for the sections of info that we probably want to add about the vaccines (as well as some possible tweaks to the vaccine figure).
However, I'm starting to think we might just want to split this section out and submit the therapeutics paper separately, since that one is basically ready to go and this needs some additions still -- plus we are already like 8000 words over the word limit 😓 

### Related issues

#738

@ajlee21 @nilswellhausen I would love to get your feedback since you have both been very involved with what's been done!
@likhithakolla and @cgreene as well, since you had written a lot of the mRNA vaccine section
And @RLordan you have been giving very helpful feedback on #740 and #741 as well, so just in case you have any thoughts!

### Checklist

### DIFF
--- a/content/20.pharmaceuticals.md
+++ b/content/20.pharmaceuticals.md
@@ -790,6 +790,8 @@ Dendritic cells are potent antigen-presenting immune cells that easily take up m
 Ultimately, primed T follicular helper cells can stimulate germinal center B cells that also present the viral antigen to produce antibodies against the virus [@doi:10.1016/j.immuni.2014.10.004].
 These cells are isolated from the patient, grown and transfected _ex vivo_, and reintroduced to the patient [@doi:10.1038/nrd.2017.243].
 
+####### ModernaTX mRNA Vaccine
+
 mRNA-1273 was the first COVID-19 vaccine to enter a phase I clinical in the United States.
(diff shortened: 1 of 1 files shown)


RLordan:
Hey @rando2, my thoughts would be that it probably makes sense to split the vaccine section up into its own manuscript if the therapeutics paper is miles over the word limit. Ultimately, @ajlee21, @nilswellhausen etc. have contributed the most so go with their thoughts.

If you do split, there is plenty updates on each of those vaccines to include. I think maybe the discussion should also briefly address the current state of vaccine distribution e.g. Moderna & Pfizer leading the way in most countries etc. mentioning the logistical issues of transport and distribution currently exposed. That would round the outline off nicely. I can help out towards the end of the week.

nilswellhausen:
I agree that it probably makes sense to split vaccines since that topic has enough volume to be its own manuscript. That way we can focus on updating the vaccine section separately and go into the data more in-depth rather than trying to squeeze this in here.

ajlee21:
Yes I very much agree with the above comments. Will start trying to take a look at the vaccine sections later this week

rando2:
Thank you so much everyone! If someone would be able to approve #740, #741, and #828, I can merge them in (we can always change the content later) and then open a new PR to split the vaccines info out! I agree this will work out for the best!

RLordan (approved):

RLordan:
> Thank you so much everyone! If someone would be able to approve #740, #741, and #828, I can merge them in (we can always change the content later) and then open a new PR to split the vaccines info out! I agree this will work out for the best!

Done!

rando2:
Thank you so much @rlordan!

PR #837, opened 2021-02-05
Quick update to favipiravir section

rando2:

This section hadn't been updated in a while -- turns out, not a lot has happened. I added in a paragraph describing some studies that have come out in August, all of which are very small. It might be worth expanding on the Russian study that found an effect on viral clearance at 4 days -- although that seems a bit odd to me given that the effect disappeared by day 10 (primary endpoint) and the study wasn't very big (60 participants total). Happy to expand on it if it would be of interest.

### Checklist

### DIFF
--- a/content/20.pharmaceuticals.md
+++ b/content/20.pharmaceuticals.md
@@ -152,7 +152,7 @@ The selection of patients did not take into consideration important factors such
 The study was neither randomized nor blinded, and the baseline control group was another antiviral instead of a placebo.
 Additionally, it should be noted that this study was temporarily retracted and then restored without an explanation [@url:https://retractionwatch.com/retracted-coronavirus-covid-19-papers/].
 
-In late 2020 and early 2021, the first randomized controlled trials of favipiravir for the treatment of COVID-19 released results [@doi:10.1016/j.ejps.2020.105631; @doi:10.1007/s00705-021-04956-9].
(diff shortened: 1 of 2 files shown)


content/20.pharmaceuticals.md
```diff
@@ -150,7 +150,20 @@ Overall, despite the study reporting clinical improvement in favipiravir-treated
...
+The sample size for this study was very small, with 29 total patients enrolled, and no significant effect of the treatments was found for the primary or any of the secondary outcomes analyzed, which included mortality.
+The second study was larger, with 96 patients enrolled, and also utilized a randomized design.
+This study enrolled only patients with mild to moderate symptoms and randomized them into two groups: one receiving CQ in addition to SOC, and the other receiving favipiravir in addition to SOC.
+This study reported a non-significant trend for patients receiving favipiravir to have a shorter hospital stay and less likelihood of progressing to mechanical ventilation or to an oxygen saturation < 90%.
+However, given the fact that favipiravir was being compared to CQ, which is now widely understood to be ineffective for treating COVID-19, these results do not suggest that favipiravir was likely to have had a strong effect on these outcomes. 
+On the other hand, another trial of 60 patients reported a significant effect of favipiravir on viral clearance at four days (a secondary endpoint analyzed) [@doi:10.1093/cid/ciaa1176].
```
cgreene, line 164: I think that you should definitely include that the effect was not observed at 10 days, especially as that was the primary endpoint.

rando2:
```suggestion
On the other hand, another trial of 60 patients reported a significant effect of favipiravir on viral clearance at four days (a secondary endpoint analyzed), but not at 10 days (the primary endpoint) [@doi:10.1093/cid/ciaa1176].
```

cgreene (approved):
Looks good! 🚀

PR #838, opened 2021-02-05
Fix broken DOI reference

agitter:
This is a valid DOI, but crosscite.org gives an invalid response when Manubot tries to fetch the metadata.  I'm switching to the PMID instead.


cgreene (approved):

cgreene:
I noticed this in my local build too - thanks!

PR #846, opened 2021-02-08
Attempt to fix scite

cgreene:
### Checklist


agitter:
I don't see the scite icons in the HTML or the PDF outputs.

cgreene:
@vincerubinetti does this look like what you expected?

vincerubinetti:
Yes, this just basically just exits the plugin before it ever inserts the scite badges if the document is in "print" mode, essentially.

vincerubinetti:
My apologies, the line should actually be:

```javascript
// if printing, exit and don't show badges
if (window.matchMedia('print').matches) return;
```

cgreene:
We've got the HTML working now. The PDF now has the loading scite badge. Should we silence that?

Screen Shot 2021-02-08 at 3 15 27 PM
Screen Shot 2021-02-08 at 3 15 24 PM

vincerubinetti:
Unfortunately that means the line I added didn't work :( That "loading scite badge" gets inserted by the code below it, and shouldn't be reached.

But I can think of no reason it wouldn't work. It may be an issue with Athena...

cgreene:
Interesting that appveyor failed... GitHub actions artifact download seems to work (that's where I screenshotted)

vincerubinetti:
I'm attempting to clone to add commits to debug this issue, but the repo is huge and my internet is having problems. If someone wants to change this line:

https://github.com/greenelab/covid19-review/blob/master/build/plugins/scite.html#L41

to:

```javascript
 badge.innerHTML = `print: ${window.matchMedia("print").matches} screen: ${window.matchMedia("screen").matches}`;
```

vincerubinetti:
Well according to my debugging, Athena does indeed think it is still a screen. I think this is an issue I can't fix. I can hide the scite badge with CSS (assuming the CSS `@media print`  is working like i think it is) which will fix the PDF in gh-actions artifacts, but the javascript will still run, and the scite fetches will still run, and appveyor will still crash.

agitter:
Another strategy would be to have the build script take different actions based on the CI service being used.  Would we be happy if GitHub Actions HTML and PDF outputs had scite and AppVeyor HTML and PDF outputs did not have scite?

Now that we cached references, we could also build the HTML manuscript twice.  The second time it would build without scite, and we could pass that version to athenapdf.

cgreene:
This seems like the most appropriate solution:
> Now that we cached references, we could also build the HTML manuscript twice. The second time it would build without scite, and we could pass that version to athenapdf.

It would add some build time... but maybe not so much.

cgreene:
Do we define a print stylesheet? Athena says this:
> Automatically falls back to screen stylesheets if no print stylesheet is defined

https://github.com/arachnys/athenapdf/blob/master/cli/README.md

It seems like we must since the PDF renders differently...

vincerubinetti:
> Do we define a print stylesheet? Athena says this:

See my comment above:

https://github.com/greenelab/covid19-review/pull/846#issuecomment-775530383

agitter:
> This seems like the most appropriate solution

I attempted that solution in d024917.  Let's see how the build artifacts turn out.

cgreene:
IT WORKS! 🚀

agitter (approved):
HTML has scite:
image

PDF does not have scite:
image

PR #847, opened 2021-02-08
Try caching reference snapshot

cgreene:

This won't automatically update the cache (at least, I don't think it will) but if someone does a local build from time to time and updates this, hopefully it'll speed up the build. 🤞 🚀 

### Checklist


agitter:
I don't think these are being used to cache references.  If you rename `output/references.json` to `content/manual-references-2021-02-08.json`, hopefully that will work.  You can delete those other two files.

This may store some bad cached reference metadata, like URL citations for which only the URL was saved.  However, we could clean that up later once the builds are working again.

cgreene:
@agitter is right. No change in build time.

Screen Shot 2021-02-08 at 12 20 28 PM
Screen Shot 2021-02-08 at 12 20 23 PM

agitter (approved):
It worked this time!  We should have done this months ago.
image

PR #819, opened 2021-01-28
Mention diet-inflammation link

rando2:
### Description of the proposed additions or changes

Hi @RLordan! I'm so sorry because I know I said I was done editing, but I couldn't sleep and was thinking about inflammation (like a totally normal person) and I thought it might be worth mentioning that diet + inflammation is not exactly an out-there concept! The compounds you talk about here are very different than the super scammy ones, like colloidal silver, for the reasons you enumerate + the fact that diet does matter.

### Checklist

- [ ] Text is formatted so that each sentence is on its own line.

- [ ] Pre-prints cited in this pull request have a GitHub issue opened so that they can be reviewed.

PR #848, opened 2021-02-09
Finalize and build pharmaceuticals manuscript

rando2:

This PR:
1. Removes a comment (based on confirming the text in the manuscript matches the reference)
2. Removes some EOL whitespace
3. Adds pharmaceuticals authorship info to metadata.yaml
4. Sets the individual docx manuscript built to pharmaceuticals

### Related issues

#738 

### Checklist


content/20.pharmaceuticals.md
```diff
@@ -269,7 +269,7 @@ Although N3 is a strong inhibitor of SARS-CoV-2 _in vitro_, its safety and effic
 After the design and confirmation of N3 as a highly potent Michael acceptor inhibitor and the identification of Mpro's structure [@doi:10.1038/s41586-020-2223-y; @doi:10.1126/science.abb3405], 10,000 compounds were screened for their _in vitro_ anti-Mpro activity.
 The six leads that were identified were ebselen, disulfiram, tideglusib, carmofur, and PX-12.
 _In vitro_ analysis revealed that Ebselen had the strongest potency in reducing the viral load in SARS-CoV-2-infected Vero cells [@doi:10.1038/s41586-020-2223-y].
-Ebselen is an organoselenium compound with anti-inflammatory and antioxidant properties [@doi:10.1007/s11033-014-3417-x] 
+Ebselen is an organoselenium compound with anti-inflammatory and antioxidant properties [@doi:10.1007/s11033-014-3417-x]
```
cgreene, line 272:
```suggestion
Ebselen is an organoselenium compound with anti-inflammatory and antioxidant properties [@doi:10.1007/s11033-014-3417-x].
```

cgreene (approved):
One missing period!

rando2:
Figures are not showing up in the docx build. I'm going to check whether this broke with the update to rootstock? I can't think of what else we've done that would have affected it!

vincerubinetti:
By the changes, you mean the plugin refactor from a couple days ago? https://github.com/manubot/rootstock/commit/87c11957cf69b73ebed5518f7928b99adb1dc5d6

All of the changes in that PR should only affect the html output, as those files only get inserted in the html build:
https://github.com/manubot/rootstock/blob/main/build/pandoc/defaults/html.yaml

I'm at a loss for why the docx would be different now.

rando2:
Thanks @vincerubinetti, it was the plugin refactor -- here it was #840 and I saw a couple of files outside the plugins directory were changed (e.g., ci/deploy.sh), so I was hoping there might be some crossover! 

I might manually fix this for now because I'm so far behind on getting this out to authors, but I'm curious if @agitter or @cgreene has any of ideas of what might have changed with the docx output? It's got something to do with rendering pictures in Word, I think, because some of them are static and some are dynamic yet none of them are showing up. We know it was still working at #819, which was the last time we built a docx with a figure in it.

vincerubinetti:
> I saw a couple of files outside the plugins directory were changed (e.g., ci/deploy.sh), so I was hoping there might be some crossover!

My specific rootstock PR didn't touch anything outside of the plugins directory, so perhaps it's been a while since this manuscript pulled from upstream and there have been other rootstock changes that I'm not aware of in that time (like `ci/deploy.sh`). You should probably ask @dhimmel and @agitter if they know anything else that changed that could've caused this problem.

agitter:
I would guess this has more to do with general issues with svg images in docx outputs than any recent rootstock changes.  There is some discussion of this in the abandoned pull request https://github.com/manubot/rootstock/pull/40  The final comments there hint that svg images in docx works on some systems but not our CI build environments.  Using png images in docx outputs is reliable.

I can help brainstorm general solutions.  However, I probably won't get to it until Thursday or Friday.  If you want to get this out to co-authors, manually fixing the images this time may be a reasonable workaround.

vincerubinetti:
Ah I didn't see that these were SVG files. I didn't think Word really supports SVGs that well. I'm surprised it ever worked at all.

content/metadata.yaml
```diff
@@ -278,6 +313,12 @@ authors:
           - Software
           - Project Administration
           - Writing - Review & Editing
+      pharmaceuticals:
+        order: 10
+        contributions:
+          - Writing - Review & Editing
+          - Project Administration
+          - Visualization
```
agitter, line 321:
```suggestion
          - Visualization
          - Software
```
Adding software because this is the section where the external resources Python scripts are used.

agitter:
I checked the docx artifact ([pharmaceuticals-manuscript.docx](https://github.com/greenelab/covid19-review/files/5954090/pharmaceuticals-manuscript.docx)), and the problem is only the svg images.  We have .png versions of those already generated.  We could switch to the .png versions instead of the .svg temporarily or permanently.

The issue with svg images in docx is a general pandoc issue: `https://github.com/jgm/pandoc/issues/4058`.  There may be a fix coming, but we shouldn't wait for it.

Figure 3 in the docx has extra whitespace at the bottom that could be cropped before the final version is submitted.

The Unicode characters don't show up correctly in the docx either.  For example, when I view the docx I see:
image

rando2:
Thanks @agitter, I just noticed that I was wrong about which figure was which! That's great that it's just the svgs. And thank you for catching my oversight on the contributions!

Whitespace issue I can address, and I actually think that the Unicode issue will be an easy fix. It seems like I mostly added the ones that were in the pathogenesis section, which means they were done with html formatting (e.g., TNF-&alpha;). I believe those all rendered correctly, or at least none of us caught issues in the docx and several of us read it closely. In this section, it seems like they are all just the unicode characters themselves. Although this will be more annoying for authors, it's good to know it's not too difficult of a fix!

rando2:
@agitter I think I fixed these issues except that I'm not sure where the specification for the file extension for the dynamic figures is provided, since the syntax in the document looks like this-- 
`**COVID-19 clinical trials.**{#fig:ebm-trials secno=1}`

Grep isn't helping me figure out which script is pulling the associated image. Do you have any thoughts on where this happens in the pipeline?

agitter:
The file extensions are hidden in too many layers of indirection to find easily.  They are being specified in .json files on the `external-resources` branch, and Manubot reads those to populate the values of the template variables in the text.

I pushed fd683829a1b58b6f2f75df678e5345e0de2b3581 to the `external-resources` branch to switch from svg to png.  This should apply automatically starting with the next build.  That change will also apply to PDF and HTML builds so we can double-check those.  If it looks bad in the HTML, I could look into using svg by default and png for docx.

cbrueffer:
Commenting here in addition to gitter, the links in references 210 and 211 seem to 404, can anyone confirm? They have a nonsensical headline ("people") as well.

agitter (approved):
I re-ran the GitHub Actions build and the .png images show up in the docx now.  They look okay in the PDF and HTML but a little pixelated.  I can try increasing the dpi later.

> Commenting here in addition to gitter, the links in references 210 and 211 (Synairgen) seem to 404, can anyone confirm? They have a nonsensical headline ("people") as well.

 I plan to do a final read through Wednesday or Thursday and will try to remember to fix this.  We can use a manual reference to set the titles and switch to the Web Archive URL.

rando2:
Thank you @cbrueffer and @agitter for catching that! I believe I was looking at press releases on their website about a week ago, so it's odd that they're already 404-ing.

I think the first one can be found here: https://www.globenewswire.com/news-release/2020/03/18/2002574/0/en/Synairgen-to-start-trial-of-SNG001-in-COVID-19-imminently.html

And the second here: https://www.icpcovid.com/sites/default/files/2020-08/200720-Synairgen-annmounces-positive-results-from-trial-of-SNG001-in-hospitalised-COVID-19-patients.pdf

I'm not sure why the links from their own website no longer work and I don't know what the best way to handle this would be.

I'm going to merge this in but I'll open an issue based on @cbrueffer's comment!

PR #858, opened 2021-02-12
Switch clinical trial analysis to interventional only

rando2:

This PR filters the `trials_df` dataframe to create a df (`interventional_trials`) that contains only the interventional trials. This was suggested by @byrdjb. It then adjusts the rest of the figure panels and the variables used in-line in the manuscript to use (or at least differentiate) these trials. 

I ran it locally but the figures haven't updated, however the json file does look like it updated. I'm not sure if there's something else I need to do to rearrange the panels in the figure? I'm ashamed to say I usually do this type of analysis in R!

### Related issues

Related PR: #859 

@agitter 

### Checklist


agitter (approved):
Your trial statistics updates make sense to me.  There was a merge conflict with the image files from when the workflow ran overnight.  I merged locally and then re-ran the scripts locally to resolve that.  Hopefully the new images will show up in the manuscript once we merge this.

> I ran it locally but the figures haven't updated, however the json file does look like it updated

The figures aren't updated in the manuscript until we merge back to the `external-resources` branch.  The build script gets the latest json files from that branch and uses those to determine which version of the figures to use.

rando2:
Hi @agitter, I'm so sorry I missed this yesterday, I haven't figured out how to replicate what you did yesterday to fix the merge issues!

agitter:
It's a pretty convoluted workflow.  Here's what I did today to resolve the merge conflicts with the external-resources branch, regenerate local figures and stats, and overwrite the latest commits on the branch.  The trick where the commit is inserted into the image URLs in a separate step makes the process especially confusing.
```
git checkout external-resources
git pull origin external-resources
git checkout external-resources-filter
git merge --strategy-option=ours external-resources
bash csse/generate-csse-stats.sh
bash ebmdatalab/generate-ebmdatalab-stats.sh
git add csse/* ebmdatalab/*
git commit -m "Update figures locally"
bash version-figures.sh
git add */*.json
git commit -m "Update stats locally"
git push -f https://github.com/rando2/covid19-review.git external-resources-filter
```

I'll merge so we don't have to resolve conflicts again.

rando2:
Thank you so much for the workflow, @agitter! I think my issue was that I didn't run version-figures.sh!

ebmdatalab/generate-ebmdatalab-stats.py
```diff
@@ -211,30 +215,30 @@ def main(args):
     plt.rc('figure', titlesize=24)
     fig, axes = plt.subplots(nrows=2, ncols=2, figsize=(20, 12), constrained_layout=True)
 
+    # Plot study type
+    # Only include study types used in >= 5 trials
+    study_type_counts = trials_df['study_type'].value_counts(ascending=True)
+    study_type_counts = study_type_counts[study_type_counts >= 5]
+    ax = study_type_counts.plot(kind='barh', ax=axes[1, 0])
```
agitter, line 222: @rando2 the panel order is hard-coded in the multi-panel figure.  It doesn't matter which order the panels are added to the figure.

Here `ax=axes[1, 0]` is the part that determines which grid position to use for the panel.  You can modify this if you want to rearrange the panels.  The `axes` list has the 0-indexed row number followed by the 0-indexed column number in the grid.

If you do want to change this, I suggest only modifying the .py file.  You don't need to update the images and json locally for such a minor change.  That will make it easier to merge, and we'll get the re-ordered panels the next time the nightly workflow runs.

PR #859, opened 2021-02-12
Update clinical trials variables/figures in manuscript

rando2:

This PR builds on #858 to switch the clinical trials figures/variables to report only interventional clinical trials, in line with a suggestion from @byrdjb.

@agitter 

### Checklist

### DIFF
--- a/content/20.pharmaceuticals.md
+++ b/content/20.pharmaceuticals.md
@@ -92,9 +92,10 @@ As results become available from additional clinical trials, we will continue to
 
 ![
 **COVID-19 clinical trials.**
-There are {{ebm_trials}} COVID-19 clinical trials and {{ebm_trials_results}} trials with results as of {{ebm_date_pretty}}.
-Interventional trials are the most frequent type of trial and were analyzed in the figures depicting status, phase, and intervention.
-The recruitment statuses and trial phases are shown only for interventional trials in which the status or phase is recorded.
+There are {{ebm_all_trials}} COVID-19 clinical trials of which {{ebm_interv_trials}} are interventional.
+Interventional trials only are analyzed in the figures depicting status, phase, and intervention.
+Of the interventional trials, {{ebm_trials_results}} trials had reported results as of {{ebm_date_pretty}}.
+Recruitment status and trial phase are shown only for interventional trials in which the status or phase is recorded.
 The common interventions are all interventions used in at least ten trials.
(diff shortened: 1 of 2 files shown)


content/20.pharmaceuticals.md
```diff
@@ -92,9 +92,10 @@ As results become available from additional clinical trials, we will continue to
 
 ![
 **COVID-19 clinical trials.**
-There are {{ebm_trials}} COVID-19 clinical trials and {{ebm_trials_results}} trials with results as of {{ebm_date_pretty}}.
-The recruitment statuses and trial phases are shown only for trials in which the status or phase is recorded.
-The study types include only types used in at least five trials.
```
agitter, line 97: Should we keep this note about which study types are shown?

agitter (changes requested):
These changes look great.  We'll have to merge #858 first to test the new template variables in the build.

Can you also please update the caption for the map in Figure 6 so we don't forget to do that later?  It looks like that also uses only interventional trials now.

content/20.pharmaceuticals.md
rando2: Yes, I agree we should! Should be fixed in 235e5e3

rando2:
Good catch! Choropleth caption is also updated now, 677ec39

agitter (approved):

agitter:
Looks great.  Now that #858 is merged we could try re-running the Appveyor build to make sure the images show up correctly before merging this.

rando2:
Oh no! I was too quick on the merge! I can revert if that's better!

agitter:
Let's leave it.  We can debug images in a new PR if we need to.

agitter:
It looks like the updated interventional figures worked: https://greenelab.github.io/covid19-review/v/ff44f6b8397192b6b47d6332beae03e66f25583d/

rando2:
@agitter I'm actually not sure if that figure was generated using the updated code -- I thought I rearranged the figure panels so that the breakdown of interventional vs other trial types was first (I tried to, at least, I haven't done too much plotting in Python recently since most of my old lab worked in R). Do we need to wait until tomorrow to see these changes in the build?

agitter:
@rando2 the subpanel order wasn't actually changed in #858.  I'll comment there in case you want to make a quick new pull request that reorders the panels.

PR #874, opened 2021-02-19
Pharmaceuticals touchups

agitter:
### Description of the proposed additions or changes

This follows up on some of the recent updates to the therapeutics manuscript, including manual references.  I changed the vaccines section cross-reference to use the HTML output instead of the Markdown source.  It is currently defined in a manual reference, but would be converted to use the arXiv preprint once that is available.

### Related issues

#863

### Suggested reviewers (optional)

@rando2 @SiminaB

### Checklist

- [x] Text is formatted so that each sentence is on its own line.

- [x] Pre-prints cited in this pull request have a GitHub issue opened so that they can be reviewed.

PR #875, opened 2021-02-20
Pharmaceuticals touchups part 2

agitter:
Follow up to #874


agitter:
There are still some problems with the URL references that have forbidden characters.  I'll take care of those before marking this as ready for review.

agitter:
The figures and manual references are working now.

rando2 (approved):

PR #894, opened 2021-03-09
Stop building docx for unneeded manuscripts

rando2:

We don't need these to build and hopefully removing them will speed up the checks a bit!

### Checklist

### DIFF
--- a/content/individual-manuscripts.txt
+++ b/content/individual-manuscripts.txt
@@ -1,2 +1,0 @@
-nutraceuticals
-pharmaceuticals


agitter:
Let's see what the build log looks like before merging.  I believe I locally tested a build where the individual manuscripts file is empty, but I'm not positive.

If there is an error, we can also set `BUILD_INDIVIDUAL` to false here
https://github.com/rando2/covid19-review/blob/feeac73227b6bba9d5dd44baeb67779dd49ee4a5/.github/workflows/manubot.yaml#L22

PR #918, opened 2021-03-30
Create appendix for vaccines

rando2:

~The original abstract, introduction, and discussion are not longer helpful to the paper and have been moved to other sections. I realized there is probably an easy way to remove them from the build without losing the history, and I confirmed that blame still shows the commit history for these files.I also may use this folder while adapting the therapeutics text to address reviewer feedback (since I am not 100% sure what the best approach to appendices will be)~

I'm just going to leave those files and move the vaccine files around right now because that's what is needed for #923 

@agitter @cgreene 

### Checklist


cgreene:
I think that you could delete the files instead of putting them in an archive folder. You would retain their history in the repository. Is the build failure expected?

rando2:
Very weird! Appveyor worked and the PDF looks fine there. I'll see if re-running fixes things here.

In terms of deleting, I've found it easier to be able to compare the commit history with blame on the original files (since I end up moving authors onto the other papers if they wrote certain sections of the intro, but it's time-consuming to match the commit history up with exact pieces of text). I'm not sure if I'm missing something, but I actually have deliberately not updated my own fork for about 6 months in order to preserve the original (single) manuscript for this reason -- I wish I'd thought it through better before restructuring!

agitter:
The way we build individual manuscripts makes strict assumptions about the filenames of the content markdown files.  That's what caused the GitHub build to fail on 5455b68.  If you need it to be more flexible and build a main text and appendix for each individual manuscript, let me know what your goal is.  I should be able to adapt it.

rando2:
@agitter The error was very clear so it was an easy fix! I just used touch to recreate an empty file, so I think this should make Manubot happy while cleaning up the diff is #923 :)

ajlee21 (approved):

Issue #906, opened 2021-03-16
Add auto-updating values and figures to Vaccines


rando2:
- Can potentially automate global vaccination rate. Will need to pick a date with a lot of data and sum. Data is in daily_vaccinations_raw column in https://github.com/owid/covid-19-data/blob/master/public/data/vaccinations/vaccinations.csv

ajlee21:
I think this might be auto-updated in another manuscript?
`SARS-CoV-2 has infected over 125 million people and cost the lives of 2.8 million people globally`

PR #922, opened 2021-03-31
Pull data from OWID to update vaccine distribution info

rando2:

See #906 
This is an implementation to pull the updated numbers so that we can add them into the manuscript.

However,  @agitter, I couldn't get the SHA hashing to work (I think it may have just been coded into the original JSON and is not available for this dataset?), so I just wanted to check whether this implementation is OK or whether there's something else that needs to happen to prevent having this run too often!

@RLordan we should be able to use this to get figures pretty easily -- I am not going to do that yet until we're positive which figs we want, but I previously wrote code to make similar maps (for the SDOH/inequality section), so I think it should be pretty straight-forward to implement building off the code here.

@ajlee21 is keeping an eye out for other things we might want to extract!

### Related issues

#906

@agitter @ajlee21 

### Checklist


owiddata/generate-owiddata-stats.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+# Get the most recent commit affecting this file and store the commit sha and date
```
rando2, line 3: Commented out because it might be important to fix it

owiddata/generate-owiddata-stats.sh
```diff
@@ -0,0 +1,16 @@
...
+#OWID_COMMIT_JSON=$(curl -sS "https://raw.githubusercontent.com/owid/covid-19-data/master/public/data/vaccinations/vaccinations.json")
+#export OWID_COMMIT_SHA=$(echo $OWID_COMMIT_JSON | python -c "import sys, json; print(json.load(sys.stdin)[0]['sha'])")
+#export OWID_COMMIT_DATE=$(echo $OWID_COMMIT_JSON | python -c "import sys, json; print(json.load(sys.stdin)[0]['commit']['author']['date'])")
+
+OWID_STATS_JSON=owiddata/owiddata-stats.json
+
+echo "Generating Our World in Data COVID-19 vaccine statistics"
+python owiddata/generate-owiddata-stats.py $OWID_STATS_JSON
+
+# After running this Python script to generate the figures, commit the figures
```
rando2, line 14: Commented out because we might be adding figures so might need to make adjustments to achieve this

owiddata/generate-owiddata-stats.py
```diff
@@ -0,0 +1,80 @@
...
+import datetime
+import json
+from datetime import date
+import matplotlib.pyplot as plt
+import os
+import pandas as pd
+#import geopandas
+#import pycountry
+
+def convert_date(git_date):
```
rando2, line 11: Copied from @agitter's ebmdatalab implementation

owiddata/generate-owiddata-stats.py
```diff
@@ -0,0 +1,80 @@
...
+
+    # Remove the leading zero of the day
+    # Assumes the year will not begin with 0
+    return datetime.datetime.fromisoformat(git_date).strftime('%B %d, %Y').replace(' 0', ' ')
+
+def main(args):
+    # Import data from github.com/owid/covid-19-data
+    owid_stats = dict()
+
+    if('OWID_COMMIT_SHA' in os.environ):
```
rando2, line 29: Doesn't do anything currently

owiddata/generate-owiddata-stats.py
```diff
@@ -0,0 +1,80 @@
...
+
+    # Identify number of vaccine manufacturers included in location totals (not the same as manufacturer-specific data)
+    owid_stats["vaccine_types"] = format(len(set([item.strip() for countryList in vaccine_locations["vaccines"].to_list() for
+                                           item in countryList.split(",")])))
+
+    with open(args.output_json, 'w') as out_file:
+        json.dump(owid_stats, out_file, indent=2, sort_keys=True)
+    print(f'Wrote {args.output_json}')
+
+if __name__ == '__main__':
```
rando2, line 71: Copied from @agitter's ebmdatalab implementation

owiddata/generate-owiddata-stats.py
```diff
@@ -0,0 +1,80 @@
...
+    # Import data from github.com/owid/covid-19-data
+    owid_stats = dict()
+
+    if('OWID_COMMIT_SHA' in os.environ):
+        owid_stats['owid_commit_sha'] = os.environ['OWID_COMMIT_SHA']
+    if('OWID_COMMIT_DATE' in os.environ):
+        owid_stats['owid_commit_date'] = os.environ['OWID_COMMIT_DATE']
+        owid_stats['owid_commit_date_pretty'] = convert_date(os.environ['OWID_COMMIT_DATE'])
+
+    # Retrieve data from OWID
```
rando2, line 35: The proposed new code is from here to the end of this function

owiddata/generate-owiddata-stats.py
```diff
@@ -0,0 +1,80 @@
...
+    locations_url = 'https://raw.githubusercontent.com/owid/covid-19-data/master/public/data/vaccinations/locations.csv'
+    vaccine_locations = pd.read_csv(locations_url, error_bad_lines=False)
+
+    numbers_url = 'https://raw.githubusercontent.com/owid/covid-19-data/master/public/data/vaccinations/vaccinations.csv'
+    vaccine_nums = pd.read_csv(numbers_url, error_bad_lines=False)
+
+    manufacturer_url = 'https://raw.githubusercontent.com/owid/covid-19-data/master/public/data/vaccinations/vaccinations-by-manufacturer.csv'
```
rando2, line 42: Don't currently use but may use for some of the visualizations

rando2 (commented):
Notes about implementation

owiddata/owiddata-stats.json
```diff
@@ -0,0 +1,7 @@
+{
```
rando2, line 1: @RLordan this is what might interest you -- the current data we're pulling from OWID.

owiddata/generate-owiddata-stats.py
```diff
@@ -0,0 +1,80 @@
+import argparse
+import datetime
+import json
+from datetime import date
+import matplotlib.pyplot as plt
+import os
+import pandas as pd
+#import geopandas
```
ajlee21, line 8: super nitpicky -- do you want to delete these if they're not being used?

ajlee21 (approved):
This looks good to me...super cool scraping!

owiddata/generate-owiddata-stats.py
rando2: Good point! I'll take them out in case we don't end up adding figures.

owiddata/owiddata-stats.json
RLordan: Super cool @rando2 , I am looking forward to this!

owiddata/generate-owiddata-stats.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+# Get the most recent commit affecting this file and store the commit sha and date
+# See https://stackoverflow.com/questions/50194241/get-when-the-file-was-last-updated-from-a-github-repository/50204589
+#OWID_COMMIT_JSON=$(curl -sS "https://raw.githubusercontent.com/owid/covid-19-data/master/public/data/vaccinations/vaccinations.json")
```
agitter, line 5: The EBM version is calling https://api.github.com as opposed to downloading the raw file.  Do you want me to try to fix this?  It may be hard for me to get the syntax correct via GitHub suggestions.  I could push to your branch to edit only these lines.

I do think we should track `OWID_COMMIT_SHA` and `OWID_COMMIT_DATE`.  These be helpful to figure out how often the external data source is updating and what version of the data we have.

owiddata/generate-owiddata-stats.py
```diff
@@ -0,0 +1,77 @@
...
+    owid_stats = dict()
+
+    if('OWID_COMMIT_SHA' in os.environ):
+        owid_stats['owid_commit_sha'] = os.environ['OWID_COMMIT_SHA']
+    if('OWID_COMMIT_DATE' in os.environ):
+        owid_stats['owid_commit_date'] = os.environ['OWID_COMMIT_DATE']
+        owid_stats['owid_commit_date_pretty'] = convert_date(os.environ['OWID_COMMIT_DATE'])
+
+    # Retrieve data from OWID
+    locations_url = 'https://raw.githubusercontent.com/owid/covid-19-data/master/public/data/vaccinations/locations.csv'
```
agitter, line 33: For `locations_url`, `numbers_url`, and `manufacturer_url` I think we should download from a specific commit instead of `master`.

I had been downloading these in the wrapper shell script, but this is slightly better to do it in the Python script if we can lock the versioning.

agitter (commented):
This is great @rando2.  Let me know how you want to proceed with versioning the data.  I can work on it if you'd like.

owiddata/generate-owiddata-stats.sh
rando2: Thank you so much @agitter, that would be awesome! Please feel free to push to the branch, or let me know if I can make it easier in any way!

owiddata/generate-owiddata-stats.py
rando2: That sounds great to me! They seem to update once per day, so it seems like we would definitely only need to rerun once per day. I'm not sure how to do this, though. I assume there's some way to ask for the most recent commit?

rando2 (commented):
Thank you @agitter for the feedback (and the awesome code base to copy off of, it made setting this up very straightforward!)

agitter (approved):
The commit versioning is working now.  This uses the GitHub API to detect the most recent commit that modified the json file we want (`vaccinations.json`).  Then it stores that commit sha and date.  The Python script then also uses that same commit to download the corresponding versions of the three csv files.  That should keep all the data in sync.

This will run nightly.  If `vaccinations.json` hasn't been modified, the script will still run but there won't be any new changes to commit.  If `vaccinations.json` was modified, the nightly run will automatically commit the updates to `owiddata-stats.json`.

I didn't check the Python code too carefully.  The structure looks good.  Assuming you manually checked the extracted data is correct, then this all looks good to me.

I also noticed that they provide a bib citation at https://ourworldindata.org/coronavirus#licence  We can use that for a manual reference when we add this to the manuscript.

rando2:
Awesome @agitter! I'm going to merge this in and we can modify as needed to get the numbers we need for the text.

PR #923, opened 2021-03-31
Vaccine organizational overhaul

rando2:
### Description of the proposed additions or changes

I'm trying to repitch this text to make it a little more dynamic for the reviewers (based on the feedback we got on therapeutics). The strategy is to contextualize & summarize here, then provide the super detailed analysis in an appendix. That's why I made two files for this manuscript for now and am copying text from the current text back over to the 22.vaccines.md in a new order. Some of it will stay in 23.vaccines-app.md (or another file) as appendices that provide greater detail about the particular vaccines.

This is very much a draft, but @ajlee21 and I were coordinating a bit, so I wanted to make sure it was out in the open.

If anyone has thoughts or wants to jump in to flesh out one of the "To Do" lines, I am more than happy to take feedback. But just keep in mind this is pretty preliminary right now! :)

### Checklist

- [ ] Text is formatted so that each sentence is on its own line.

- [ ] Pre-prints cited in this pull request have a GitHub issue opened so that they can be reviewed.

PR #832, opened 2021-02-03
New text on Colchicine

athril:
With the recent finding from COLCORONA clinical trial, I would suggest adding this drug to the manuscript 
Could you possibly review this section @rando2 @agitter ? #831

### Checklist

- [x] Text is formatted so that each sentence is on its own line.

- [ ] Pre-prints cited in this pull request have a GitHub issue opened so that they can be reviewed.

PR #757, opened 2021-01-13
move anakinra to pharmaceuticals

rando2:
### Description of the proposed additions or changes

Replaces #605 
Due to the restructure + not having the necessary write permissions on other fork, I think this is the easiest way to get a version of #605 that can be incorporated. There was great feedback on #605 so we should keep an eye on it!

### Checklist

- [ ] Text is formatted so that each sentence is on its own line.

- [ ] Pre-prints cited in this pull request have a GitHub issue opened so that they can be reviewed.

PR #927, opened 2021-04-02
Move existing therapeutics text to appendix

rando2:

This seemed to work for vaccines (see #918 and #923), so just doing the same thing to therapeutics here

### Checklist


agitter (changes requested):
The build logs show that the appendix will be ordered before the main text
```
Manuscript content parts:
00.front-matter
01.abstract
02.introduction
07.pathogenesis
09.evolution
10.diagnostics
20.pharmaceuticals-app
20.pharmaceuticals
```

The most reliable approach would be to renumber the markdown files so that we have `21.pharmaceuticals-app` and then increment the others in the 20s as well.

The good news is that I looked at the build script again, and I don't think it needs to be modified to support this new file naming convention for the main text and appendix.  As long as the main file matches the pattern `*.<keyword>.md` and the appendix markdown file contains that same keyword, it should be fine.  I'm testing it locally.

#757 and #832 won't be mergeable after this.  We could close them and decide later if we need to cover those therapeutics.  In my opinion, we don't need to in the restructured format.

rando2:
> The build logs show that the appendix will be ordered before the main text
> 
> ```
> Manuscript content parts:
> 00.front-matter
> 01.abstract
> 02.introduction
> 07.pathogenesis
> 09.evolution
> 10.diagnostics
> 20.pharmaceuticals-app
> 20.pharmaceuticals
> ```
> 
> The most reliable approach would be to renumber the markdown files so that we have `21.pharmaceuticals-app` and then increment the others in the 20s as well.
> 
> The good news is that I looked at the build script again, and I don't think it needs to be modified to support this new file naming convention for the main text and appendix. As long as the main file matches the pattern `*.<keyword>.md` and the appendix markdown file contains that same keyword, it should be fine. I'm testing it locally.
> 
> #757 and #832 won't be mergeable after this. We could close them and decide later if we need to cover those therapeutics. In my opinion, we don't need to in the restructured format.

Thank you so much for catching this @agitter. I'm not sure how we want to handle the appendices long-term but for now, this seems to work well enough! I think I made the changes needed to not break anything, and I can close those stale PRs.

agitter (approved):

PR #928, opened 2021-04-02
Make OWID stats available in manuscript

agitter:
I forgot this was needed in #926 to allow the template variables to be used in the manuscript text.


rando2 (approved):
Awesome! I can implement a test once we get a couple of PRs merged in (I'm hoping today!)

PR #931, opened 2021-04-07
Update ORCID formatting

agitter:

Formatting recently added metadata

### Checklist

### DIFF
--- a/content/metadata.yaml
+++ b/content/metadata.yaml
@@ -1109,7 +1109,7 @@ authors:
     github: rishirajgoel
     name: Rishi Raj Goel
     initials: RRG
-    orcid: https://orcid.org/0000-0003-1715-5191
+    orcid: 0000-0003-1715-5191
     email: rishi.goel@pennmedicine.upenn.edu
     twitter: rishirajgoel
     code of conduct:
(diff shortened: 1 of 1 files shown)


rando2:
Thanks @agitter! I think I made this mistake once and then just kept replicating my own error 🤦‍♀️

rando2 (approved):

agitter:
It's something that Manubot should report or fix automatically.  That could be a small feature to add in the next release.

rando2:
> It's something that Manubot should report or fix automatically. That could be a small feature to add in the next release.

Is there a good place to record these? (e.g., issue in rootstock?) I have a couple ideas as well I think just require some regex but might help with smoothing over some details in a big project like this! (Otherwise I'll just add this to my local list!)

agitter:
This particular feature idea can be an issue in https://github.com/manubot/manubot.  It would require updating that Python package to reformat or check validity of author metadata when it is read in.

Other broader ideas for the Manubot workflow can be issues in https://github.com/manubot/rootstock

We can transfer issues so it isn't a big deal to decide which repo is best.  It's very helpful to have bug reports and feature requests logged as issues so please feel free to add some.

PR #941, opened 2021-04-27
Initial plotting of CORD-19 dataset

rando2:

This is a little rough but I wanted to get it online before I go teach for 4 hours!
@mprobson @agitter I'm not quite sure what to do because Kaggle uses an API and I'm not sure how to set that up with open source software, should we just set up a project account and put the API key in the repository? Currently the biggest gap in this push is that it doesn't have a way to actually download the data!
I use two files: metadata.README and metadata.csv.zip that you can find here: https://www.kaggle.com/allen-institute-for-ai/CORD-19-research-challenge

The `.sh` file definitely needs some cleaning up and I would like to go through the documentation in the `.py` file and make sure it's correct/helpful, but I figured getting something was the first priority.

@agitter @mprobson 

### Checklist


mprobson:
I _think_ this is basically the same data but we can just grab it using `wget` or `curl` : https://ai2-semanticscholar-cord-19.s3-us-west-2.amazonaws.com/historical_releases.html . I downloaded the latest repo on my computer and it seems to contain the following:

- metadata.csv
- document_parses/
- cord19_embeddings/ 
- changelog

rando2:
Great find, @mprobson. We should be able to do this pretty easily, possibly iterating backwards from today's day and then checking the sha? I need to think about it but this site makes it WAY easier.
The files we would need can be obtained from URLs structured like:
https://ai2-semanticscholar-cord-19.s3-us-west-2.amazonaws.com/2021-04-26/metadata.csv
https://ai2-semanticscholar-cord-19.s3-us-west-2.amazonaws.com/2021-04-26/changelog

agitter:
This will be a great addition to prove the point about the explosion of literature.  I can help integrate it into the nightly build when we're happy with how it runs locally.

I initially thought it would be nice to put total manuscripts and preprints in the same figure.  However, the scale is so different that wouldn't work well, right?

rando2:
@agitter I thought the same thing! Originally I counted anything with "rxiv" into a single line on the first graph. It didn't look nearly as impressive as it is because of the scale. There are certainly other preprints I'm missing with that heuristic, though, but I couldn't think of a quick way to grab them out based on the info provided in the CORD-19 dataset.

rando2:
Updated to script pulling the data from the link @mprobson sent. There is probably a more elegant way to do this, but I hate text manipulation in bash, so I just joined the date and sha1 together and then split them in Python. 

Now I just need to look into versioning the figures and then I think this might be ready? Just default matplotlib plots but that seems fine given our timeline 😓

agitter:
>  Just default matplotlib plots but that seems fine given our timeline 😓

I agree we don't have time to perfect the figures.  We could tweak them more post-submissions.  Is it quick and easy to make the lines thicker and the text larger?

rando2:
@agitter I'm a little ashamed to admit this but I have done most of my plotting in R lately and I am super rusty on matplotlib. @mprobson has said he may have some time to take a look this afternoon and try to spruce them up a bit!

agitter:
I had a surprise meeting cancellation!  I'll push a couple small changes to integrate this with the external resources updates workflow and the manuscript.

agitter (approved):
I can run the scripts locally, and they work well after the updates I pushed.  I borrowed the matplotlib formatting style that Ryan added to the CSSE figure.  I also renamed the template variables so they all have a `cord19_` prefix.

I have not reviewed the data parsing logic carefully.  We could leave this open if someone wants to double-check that part, or we can merge now.  I'll make a separate pull request to update the build script and workflow.

agitter:
The figure with the CSSE style:

cord19-growth

rando2:
Thank you so much @agitter! Those look better. I talked to @cgreene and he was a little surprised that they have half a million papers in the corpus, but I double-checked and that is what their metadata says. CoronaCentral (which uses this dataset but curates it) has about 150,015 coronavirus papers total ([scroll to bottom](https://coronacentral.ai/)). I confirmed there are 544K lines of data (which is what the changelog said), but I'm not exactly sure what the extra data are. We could just add in context based on their paper to explain what criteria they used and cite the CoronaCentral number as well.

agitter:
> We could just add in context based on their paper to explain what criteria they used and cite the CoronaCentral number as well.

I agree with both of those ideas because half a million does seem huge.

rando2:
Excellent, I'm going to merge this in and get started on updating the text!

PR #944, opened 2021-04-28
Adapting Daniel's code for contributor graphs

rando2:

This still needs some work, but it generates something resembling the graphs that Daniel made for the Deep Review. There are a lot of compatibility issues that I'm whack-a-mole-ing but hopefully in the homestretch now. One figure worked, the other two there is some ggplot error error I'll work on tomorrow.

Also just some silly stuff that needs to be cleaned up, like I didn't update the output file names and I can't remember a lot of people's github username to real name mappings, so some of those are very very wrong (like just people's usernames a second time!)

### Checklist


rando2:
OK! This can now create two figures and the stats file about the commits. I was trying to get a third figure to show contribution by career stage, but I'm going to give up on that for now because it's non-trivial and doesn't seem like something we completely need. There is some issue with one of the date packages that is causing a lot of issues and made it so that third graph really isn't usable, regardless (I was hoping if I switched to career stage, it would be more readable, but I don't think I have time for the data manipulation required to do that!)

Here are the figs that worked for easy viewing:
covid19-review-contribution-dot
covid19-review-contribution-ridge

cbrueffer:
Very cool! I think the colors in figure 1 may be a bit hard to interpret. It's looks like most people only deleted code, when in reality I suppose the number of additions/deletions is equal, and red it just plotted last? Maybe replacing each circle with a small bar plot may work, although no idea how hard to implement.

If this is either/or between the two plots, I think the first one is more expressive.

One small nit, maybe Ashwin N. Skelly (dot after N) for consistency.

rando2:
Quick update, based on convo with Casey I changed to log scale -- this does look better but we would need to figure out how to toggle the y-axis scaling
covid19-review-contribution-ridge

agitter:
> It's looks like most people only deleted code, when in reality I suppose the number of additions/deletions is equal, and red it just plotted last?

Yes, that's what's happening.  If contributors primarily edit existing lines, those edited lines count as both added and deleted in the commit.  One solution would be to plot additions only and caption the figure appropriately to describe that as edited and added lines.

I liked the ridge plot in the original deep review summary, but I like the dot plot more here.  Even with the log scaling, the ridge plot has a lot of lines that look mostly horizontal.  I think the dot plot more accurately reflects the activity on the manuscript.  There was a big burst in spring 2020 that established the foundation.  Then, the number of active contributors diminished over time, but several contributors remained active editing the text.

I'm curious what my massive deletion was in August 2020 😄

rando2:
@cgreene @cbrueffer @agitter Here is a plot of just additions! 
covid19-review-contribution-dot

agitter:
I like it!

cbrueffer:
Nice! I agree that the number of deletions is probably not so interesting here, so leaving them out is a good option.

rando2:
@agitter I think this code would now work to be integrated with our workflow. I added a `.sh` script to run the notebooks (adapted from Daniel's Deep Review notebooks) and also added these figures to the versioning. However, I have generally only used jupyter notebooks for teaching, so had to look up a lot of these solutions -- I'm happy to change anything that could be handled better.

The one trick here is that it won't work in the base covid19-review environment, so I think I will need to also modify `.github/workflows/update-external-resources.yaml` to run these scripts first in a separate environment, then switch to the normal external-resources environment to run the rest of the figure scripts and then version everything. [This page](https://sionwilliams.com/posts/2020-02-04_jupyter_ci/) does seem to outline a solution to a similar problem, but it seems like more of an overhaul than we probably want since this is a one-off at present.

contrib-viz/01.contrib-git.ipynb
```diff
@@ -0,0 +1,591 @@
...
+       "35"
+      ]
+     },
+     "execution_count": 6,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "path = pathlib.Path('renamer.tsv')\n",
```
ajlee21, line 345: What are you manually renaming here?

contrib-viz/01.contrib-git.ipynb
```diff
@@ -0,0 +1,591 @@
...
+    "    for key in 'words_added', 'words_deleted', 'characters_added', 'characters_deleted':\n",
+    "        delta[key] = 0\n",
+    "    for diff in diffs:\n",
+    "        diff_str = str(diff)\n",
+    "        additions = added_pattern.findall(diff_str)\n",
+    "        deletions = deleted_pattern.findall(diff_str)\n",
+    "\n",
+    "        # Attempt to ignore relocated lines\n",
+    "        additions, deletions = (\n",
+    "            [x for x in additions if get_word_count(x) < 4 or x not in set(deletions)],\n",
```
ajlee21, line 91: Why less than 4?

ajlee21 (approved):
Really cool analysis and visualization! I'm not sure if I can follow all the scaping calculations performed, but the everything seems reasonable.

I know you log-scaled the values, did you also log-scale the size of the circle marker too?

Just to clarify, is this a visual meant to help inform authorship order? If so, I wonder if there are additional metrics that would be useful like "number of reviews/comments" in addition to text edits?

contrib-viz/01.contrib-git.ipynb
rando2:
```suggestion
    # Update any users whose GitHub names don't match their authorship names
    # Use information specified in tab-delim file renamer.tsv
    "path = pathlib.Path('renamer.tsv')\n",
```

contrib-viz/01.contrib-git.ipynb
rando2: Great question! This is mostly Daniel's code. I'll play around with it and figure it out!

agitter: That heuristic originated in https://github.com/greenelab/meta-review/commit/3aa682608566d55daecf0051af6c1764d008d3c7  Daniel's commit message acknowledges it is not very precise:
> This method aims to ignore lines that were relocated in the manuscript.
It is not very advanced and probably doesn't work very well.

contrib-viz/02.contrib-viz.ipynb
```diff
@@ -0,0 +1,819 @@
...
+    "options(repr.plot.width=width, repr.plot.height=height)\n",
+    "\n",
+    "gg = month_df %>%\n",
+    "  dplyr::filter(diff_type %in% c('words_added', 'words_deleted')) %>%\n",
+    "  dplyr::arrange(desc(count), desc(diff_type)) %>%\n",
+    "  ggplot2::ggplot(ggplot2::aes(x=year_month, y = username, size = count, color=diff_type)) +\n",
+    "  ggplot2::geom_point(stroke=1) +\n",
+    "  #ggplot2::scale_y_discrete(limits=usage_df$username, position = 'top', labels=y_labels, name=NULL) #+\n",
+    "  ggplot2::scale_x_discrete(limits = year_month_breaks, labels=year_month_labeler, name=NULL) +\n",
+    "  ggplot2::scale_size_area(max_size = 3.8, name='word count') +\n",
```
agitter, line 651: Because we are only showing additions, we could remove the green circle and `added` and change `word count` to `words added`.  I agree with @ajlee21's suggestion to log-scale the size of the circle marker if we can.

contrib-viz/03.contrib-stats.ipynb
```diff
@@ -0,0 +1,481 @@
...
+     ]
+    }
+   ],
+   "source": [
```
agitter, line 216: Do we need this block?  We already have different ways to track the number of authors that might conflict with this count.

contrib-viz/03.contrib-stats.ipynb
```diff
@@ -0,0 +1,481 @@
...
+      "{\n",
+      "  \"creation_time_utc\": \"2021-08-27T21:21:39.116795\",\n",
+      "  \"creation_date_pretty\": \"August 27, 2021\",\n",
+      "  \"commit\": \"2c88132aa6e259244b474470bfe72722f89f1683\",\n",
+      "  \"total_commits\": 4876,\n",
+      "  \"manuscript_commits\": 2775,\n",
+      "  \"review_authors\": 52,\n",
+      "  \"github_stars\": 112,\n",
+      "  \"github_forks\": 75,\n",
+      "  \"merged_pull_requests\": 559,\n",
```
agitter, line 408: Do we want to include `merged_pull_requests` and the number of issues in the methods manuscript?  That's nice to show how active the project has been.

contrib-viz/calc-contrib-stats.sh
```diff
@@ -0,0 +1,13 @@
...
+# and run the version-figures.sh script to update the CONTRIB_STATS_JSON with the
+# versioned figure URL
+
+echo "Generating contributor statistics and figure"
+jupyter nbconvert --to notebook --execute --ExecutePreprocessor.kernel_name=python3 --output output.ipynb contrib-viz/01.contrib-git.ipynb
+jupyter nbconvert --to notebook --execute --ExecutePreprocessor.kernel_name=ir --output output.ipynb contrib-viz/02.contrib-viz.ipynb
+jupyter nbconvert --to notebook --execute --ExecutePreprocessor.kernel_name=python3 --output output.ipynb contrib-viz/03.contrib-stats.ipynb
+
+# Remove temporary files
+rm contrib-viz/commits.tsv contrib-viz/contributors.tsv
```
agitter, line 13: Do the above commands also create `output.ipynb` that can be removed?  Or do we want to give the outputs unique names and store them?

version-figures.sh
```diff
@@ -10,6 +10,7 @@ echo "Using figure versions from commit $FIGURE_COMMIT_SHA"
 CSSE_STATS_JSON=csse/csse-stats.json
 EBM_STATS_JSON=ebmdatalab/ebmdatalab-stats.json
 CORD19_STATS_JSON=CORD-19/cord19-stats.json
+CONTRIB_STATS_JSON=contrib-viz/covid19-review-stats.json
```
agitter, line 13: I don't think this file needs to be modified yet.  We don't have the figure URLs in `covid19-review-stats.json` so they don't need to have the commit replaced in the URLs.  We could save that set of changes for another pull request that integrates this into the daily or weekly workflow.

agitter (changes requested):
This is looking good.  I have a few small suggestions, but it is nearly ready to merge.  I reviewed the main analysis code lightly because it is derived from Daniel's code that we used successfully in another project.

contrib-viz/02.contrib-viz.ipynb
rando2: These suggestions are done in the new version I'll push. But log scaling did not seem to work, it made the dots look very uniform. I agree it would be a more intuitive measure but can't seem to get it to visualize well.

contrib-viz/03.contrib-stats.ipynb
rando2: Removed!

contrib-viz/03.contrib-stats.ipynb
rando2: That's great idea! Added a sentence to the conclusion in #1033

contrib-viz/calc-contrib-stats.sh
rando2: We can remove it, I think. Everything we need is saved as a json or image file.

```suggestion
rm contrib-viz/commits.tsv contrib-viz/contributors.tsv output.ipynb
```

version-figures.sh
rando2: OK that sounds great!

rando2:
@agitter here is the updated figure:
image

agitter (approved):
Look great!  Let's merge it and add it.

I will also work on adding this to a workflow at some point post-submission.

We could probably remove "username" from the y-axis label on the dot plot now that we have full names.  Let's leave it for now.  I can manually crop it out if we want to remove it for the DISCO submission.

agitter:
I merged this so I can test it within the manuscript.

PR #946, opened 2021-04-28
Add CORD-19 resource to workflow and build script

agitter:
This follows up on #941 so that the script to generate CORD-19 statistics and figures is run daily.  It also tests that the statistics are available in the manuscript.


agitter:
The new stats are working
image

rando2 (approved):
Looks great! I'll update #947 to incorporate this!

PR #943, opened 2021-04-27
Add LaTeX export for ACM-BCB

agitter:
This adds preliminary LaTeX export for the ACM-BCB submission.  The big picture is that we can automatically generate a base document for the submission, but it will require a fair amount of manual editing before it will build with the ACM sigconf proceedings template.  Because of our tight timeline, we'll need to decide how much more to automate versus prioritizing manuscript content.

There is now a list `individual-docx-manuscripts.txt` of manuscripts to export as docx and another list `individual-latex-manuscripts.txt` to export for LaTex.  Exporting for LaTeX generates a .tex and a .bib file that contains reference metadata from Manubot.  Hopefully those two outputs are work together.  I haven't tried building a PDF yet.

This builds on two experimental rootstock pull requests: https://github.com/manubot/rootstock/pull/384 and https://github.com/manubot/rootstock/pull/386  One of those upgrades the version of Pandoc.  That will cause some changes to our HTML and PDF outputs, so we'll need to check those carefully before merging.  I also had to further update the environment to resolve new package incompatibilities.  However, the newer version of Pandoc is needed to extract the .bib file.

When outputting .tex files, Pandox uses a LaTeX template.  This is different from what ACM refers to as a template, which is an example .tex file.  The Pandoc version of a template can access metadata variables, like our Markdown template for the front matter.  This strategy isn't compatible with how Manubot writes author information and some other metadata.  I can resolve some of these incompatibilties, but we'll need to decide what missing metadata (authors? affiliations?) is high priority.

I didn't test images.  Those will probably not work immediately.

`@misc` URL citations currently don't include the URL in the bib file.

Here are the current outputs (.txt added so GitHub will allow the attachment):
- [methods-manuscript.tex.txt](https://github.com/greenelab/covid19-review/files/6388056/methods-manuscript.tex.txt)
- [methods.bib.txt](https://github.com/greenelab/covid19-review/files/6388057/methods.bib.txt)

PR #955, opened 2021-04-30
Initial add of dhimmel to authors

rando2:
@mprobson do you need conflict of interest info for ACM? I'm waiting to hear back on that but this should be most of the required metadata for submitting I think!

### Checklist


rando2:
Manubot is failing because of missing COI info!

mprobson:
> do you need conflict of interest info for ACM?

The best I could find was discussed here: https://github.com/greenelab/covid19-review/pull/943#issuecomment-830218914. Basically significant CoI should be disclosed in an unnumbered footnote on the first page. @agitter is working on automatically generating a string that is pased as metadata to the template and inserted.

agitter:
We'll need to add the structured ACM affiliation metadata for Daniel as well (following the format in #943).

We can also remove Vince and Daniel from the general acknowledgements once they are both added as authors on the merged manuscript.

PR #957, opened 2021-05-11
Methods tex edits

agitter:
I created a diff of the ACM-BCB tex file in a0fb364e7e7134afae36adb3a2850801da72735b and then manually reapplied those changes to the markdown version.  This content doesn't need to be reviewed.  The goal is to update the markdown version to match the conference submission so we can tag it and continue editing.


rando2 (approved):
Thank you so much for doing this @agitter! I was having stress dreams about how to do it in Python for the first couple nights after we submitted the paper.

PR #966, opened 2021-05-21
Modify CORD-19 script to exit with 0

agitter:
The external resources build [is failing](https://github.com/greenelab/covid19-review/runs/2625662310?check_suite_focus=true) because `CORD-19/generate-cord19-stats.sh` exits with 1.  This should allow it to ignore missing files and still exit with 0.


agitter:
This worked as intended.  The latest [build](https://github.com/greenelab/covid19-review/runs/2698944015?check_suite_focus=true) ran successfully even though the CORD-19 data was up to date.

PR #972, opened 2021-06-11
Update rootstock on 2021-06-11

agitter:
We updated to Pandoc 2.12 to support LaTeX export.  However, that broke some Manubot formatting.  Rootstock now supports the latest Pandoc, so this will fix some of the minor issues introduced with our prior Pandoc update.


rando2 (approved):
This looks reasonable to me but I didn't go through the code carefully!

PR #982, opened 2021-08-03
Update COI for @marouenbg

marouenbg:


rando2 (approved):
Thank you so much for doing this!

content/metadata.yaml
```diff
@@ -808,7 +808,7 @@ authors:
       confirmed: yes
     coi:
       string: "None"
-      lastapproved: !!str 2020-11-11
+      lastapproved: !!str 2020-08-02
```
rando2, line 811:
```suggestion
      lastapproved: !!str 2021-08-02
```
I also missed the year the first time I looked at it...

marouenbg: Thanks @rando2 !

PR #994, opened 2021-08-11
Move content from old, "one big paper" format into a separate folder

rando2:

There are some issues created by figures being added twice -- this should let us see what happens when the duplicates are removed from the build!

### Checklist


agitter:
I've lost track of what content from these old files has been copied and what might still be relevant.  Is the issue the duplicate figures, that the text in these files is no longer accurate/relevant, or something else?

I'm wondering whether we want to completely remove these sections or use templating to show them conditionally.

rando2:
@agitter I believe everything has been spread across the various papers! My only concern is that it can be helpful to have access to the "blame" function for confirming that the authorship lists are complete. I don't think they need to be rendered at all, I'm just hesitant to remove them.

We probably do want to add a generic abstract & introduction that will provide context for the HTML manuscript-as-a-whole rendering, but I haven't gotten around to that yet!

agitter (approved):
> I believe everything has been spread across the various papers

In that case, this reorganization makes sense.  With the LaTeX manuscript export we've gotten more sophisticated with selectively commenting certain sections, so I wanted to see whether that would be helpful here.  Sounds like it isn't necessary.

PR #998, opened 2021-08-12
Add a dictionary item with special characters.

cbrueffer:

The spellcheck fails on entries with special characters (e.g. hyphens), this PR demonstrates the issue.

@agitter @rando2 

### Checklist


cbrueffer:
Yes, that's fair enough. The entry in question also wasn't super critical.

PR #996, opened 2021-08-11
Edits to main therapeutics manuscript

SiminaB:
### Description of the proposed additions or changes

I made some edits to the main therapeutics manuscript.  I included a few questions as comments. Note that some of the text was just moved around, not deleted, in a way that I thought was more logical/straight to the point (feel free to disagree, of course!)

### Suggested reviewers (optional)

@rando2 @agitter  @cgreene 

### Checklist

- [x] Text is formatted so that each sentence is on its own line.

- [x] Pre-prints cited in this pull request have a GitHub issue opened so that they can be reviewed.

PR #999, opened 2021-08-12
Add x-tick labels to CSSE figure

rando2:

Per the suggestion in #996, this add in x-axis labels for the CSSE figure.
I also adjusted the matplotlib code to make it easier to apply changes to all panels

Here is the new figure:
csse-deaths

### Checklist


agitter (approved):
Great update.  The figure looks better, and the code is more organized as well.

PR #953, opened 2021-04-30
Add visualization of manuscript stats

rando2:
### Description of the proposed additions or changes

I'm not sure how it will work to have figure/statistic generation on the output branch, but I'm not totally sure if we need to worry about that today. I can just add the figure to the manuscript as a static .png file and fill in the stats for today's goals!

### Related issues

Closes #952 

### Suggested reviewers (optional)

@agitter 

### Checklist

- [ ] Text is formatted so that each sentence is on its own line.

- [ ] Pre-prints cited in this pull request have a GitHub issue opened so that they can be reviewed.

Issue #952, opened 2021-04-29
Plotting manuscript progression for methods manuscript


agitter:
For the ACM-BCB submission we could plot the following manuscript statistics over time:
- Number of authors who added themselves to metadata
- Number of references
- Word count

All of these are available in the files `variables.json` and `references.json` in the output branch of the repo.  Some quick Python experimentation shows how to access these values and the corresponding date:
```python
import json
variables = json.load(open('variables.json'))
>>> variables['pandoc']['date-meta']
'2021-04-28'
>>> variables['manubot']['date']
'April 28, 2021'
>>> len(variables['manubot']['authors'])
49
>>> variables['manubot']['manuscript_stats']['word_count']
131943
references = json.load(open('references.json', encoding='utf-8'))
>>> len(references)
1428
```

I don't know the most efficient way to get these for every commit in the output branch.  However
```
git log --pretty=format:"%h" > output-commits.txt
```
dumps a list of all commits to a text file that we could iterate over.

Pseudocode for an algorithm could look like:
- dump all commits on the output branch
- foreach output branch commit
- checkout the commit in a subprocess
- load the json files as shown above and store the commit sha1, manuscript date, author count, word count, and reference count in a row in a dataframe
- plot the data from the dataframe

Doing this with a Python script would be messy due to the subprocess calls to issue git commands, but it's possible and I don't know the GitPython package well enough to do it that way.  For example
```
subprocess.run(["git", "checkout", "3839cc2e"])
```
will checkout a specific commit from the output branch.

rando2:
Thank you so much @agitter! I'm going to try to finish getting the text cleaned up (ish), then send it to authors and start working on parsing #17, and then hopefully take this on in the morning!

rando2:
Initial prototype is working 👍 Thanks for the point in the right direction, @agitter, those json suggestions make it WAY easier than what I was imagining (which involved a lot of regex).

This is obviously EXTREMELY ROUGH and needs to be visually cleaned up in basically every way, but it *is* a graph of the data!
<img width="475" alt="Screen Shot 2021-04-30 at 8 16 13 AM" src="https://user-images.githubusercontent.com/8975841/116693784-616d4a00-a98c-11eb-9783-ae0a97573b85.png">

agitter:
That's amazing!

We should be able to flip the order of the dates and show fewer x-axis ticks (e.g. monthly) without too much trouble.

[...]

PR #1019, opened 2021-08-30
Visualize manuscript/project growth

rando2:

This is an updated version of #953 that works off of `master` to pull the commit history and then visualize commits

### Related issues

#953 #952 

### Checklist


rando2:
@mprobson this takes a very long time to run, but I think it is appropriate for map reduce. The input is a list of commits (one per line in a .txt file, generated by `git log` on the command line) and it extracts various data, then stores in a dictionary. I don't see why we couldn't merge together a bunch of dictionaries. But since it's been a while since I parallelized anything in python, figured I should run this by you!

analyze-ms-stats/calc-manuscript-stats.py
```diff
@@ -0,0 +1,86 @@
...
+    manuscript_stats = growthdata.iloc[0].to_dict()
+    for item in ["Authors", "Word Count", "References"]:
+        manuscript_stats[item] = str(manuscript_stats[item])
+    growthdata = growthdata[::-1]
+    growthdata = growthdata.set_index("Date")
+
+    # Plot the data
+    axes = growthdata.plot(kind='line', linewidth=2, subplots=True)
+    for ax in axes:
+        ax.set_ylabel('Count')
```
agitter, line 51: For the word count, this will need to be `Count (thousands)` or something similar.

analyze-ms-stats/calc-manuscript-stats.py
```diff
@@ -0,0 +1,86 @@
...
+        ax.get_yaxis().set_major_formatter(matplotlib.ticker.FuncFormatter(
+            lambda x, p: format(int(x), ',')))
+        ax.set_ylim(bottom=0)
+        ax.spines['top'].set_visible(False)
+        ax.spines['right'].set_visible(False)
+        ax.minorticks_off()
+        ax.grid(color="lightgray")
+
+    ax.figure.savefig(args.output_figure + '.png', dpi=300, bbox_inches = "tight")
+    ax.figure.savefig(args.output_figure + '.svg', bbox_inches = "tight")
```
agitter, line 61: Is the svg being written but not yet commit to this branch?

agitter (commented):
How slow is this to run?  If we need to speed it up substantially, could we write the statistics to disk for commits that have already been analyzed and only update that file with newer commits?  We would need to weigh the tradeoffs of adding that complexity to the code versus just waiting for it it run.

Do you envision this being run manually, with every new commit, or on a regular (e.g. daily/weekly) schedule?  Ideally it could auto-update in some form.

rando2:
> How slow is this to run? If we need to speed it up substantially, could we write the statistics to disk for commits that have already been analyzed and only update that file with newer commits? We would need to weigh the tradeoffs of adding that complexity to the code versus just waiting for it it run.
> 
> Do you envision this being run manually, with every new commit, or on a regular (e.g. daily/weekly) schedule? Ideally it could auto-update in some form.

Haha Tony, we're on the same page -- I was about to make a comment asking whether it would make sense to write to disk, and then realized I should just do it, since that's probably the best/only way to speed things up. It's running in a not-unreasonable amount of time with 8 CPUs, but when I had it single threaded, it was slow enough to be pretty annoying, so I definitely don't want it clogging up builds.

I don't think it needs to be run with every commit. Definitely daily and even weekly would probably be fine, since it's not really critical to have a super up-to-date record of how many references are in the document etc.

One thing I wanted to ask you is: presumably I should set up a separate conda environment because it's on a different branch from the rest of the visualizations?

agitter:
> One thing I wanted to ask you is: presumably I should set up a separate conda environment because it's on a different branch from the rest of the visualizations?

I see this relates to your comments in #944 (which I'm still reviewing).  I don't see any harm in creating separate conda environments.  If the environments are very similar, perhaps there would be some wasted time (~minutes) configuring multiple environments when one would suffice.

I'm wondering whether we should set up this analysis and #944 on the `external-resources` branch.  They aren't actually external.  However, that would keep our automated analyses more organized.  We could have a new daily/weekly workflow modeled after `update-external-resources.yaml` that updates these internal project statistics (`update-project-statistics.yaml`?).  We could also use that branch to store the latest statistics written to disk from this analysis and read from that file if it exists.  However, if we are only running this ~weekly in a GitHub Actions workflow instead of with each build, it doesn't really matter if it runs for 10s of minutes.

Do you see any pros or cons of adding these project statistics analysis to the `external-resources` branch?

rando2 (commented):
I changed it to just use the absolute word count instead of dividing the word count by 1000 (I'm sure we can get this working that way if we feel strongly that we want it, but this was easier).

In terms of run time, it takes 8-10 seconds on 12 CPUs (locally) to analyze all 498 commits. With the caching, it takes 1.2 seconds to process 10 new commits locally and 2.2 seconds/10 records if I set it to a single CPU.

rando2:
Update: to crunch everything through 8/27 took about 10 seconds using 12 CPUs


> > One thing I wanted to ask you is: presumably I should set up a separate conda environment because it's on a different branch from the rest of the visualizations?
> 
> I see this relates to your comments in #944 (which I'm still reviewing). I don't see any harm in creating separate conda environments. If the environments are very similar, perhaps there would be some wasted time (~minutes) configuring multiple environments when one would suffice.
> 
> I'm wondering whether we should set up this analysis and #944 on the `external-resources` branch. They aren't actually external. However, that would keep our automated analyses more organized. We could have a new daily/weekly workflow modeled after `update-external-resources.yaml` that updates these internal project statistics (`update-project-statistics.yaml`?). We could also use that branch to store the latest statistics written to disk from this analysis and read from that file if it exists. However, if we are only running this ~weekly in a GitHub Actions workflow instead of with each build, it doesn't really matter if it runs for 10s of minutes.
> 
> Do you see any pros or cons of adding these project statistics analysis to the `external-resources` branch?

I actually think it would be kind of nice to have all of the code-based visualization happening in one place (or at least, not filling up the part of the repository that non-coders might be looking at). For that reason, I'd be for moving them to external-resources even though they're not external!

I implemented the caching because it was bothering me. The plan to add `.yaml` files sounds good to me! I may de-prioritize that for the next couple days and try to get some more revisions done for the camera ready, but that seems like a great solution for keeping things up-to-date to me!

analyze-ms-stats/calc-manuscript-stats.py
```diff
@@ -0,0 +1,126 @@
...
+
+        if len(commits) > len(oldCommits):
+            start_old = commits.index(oldCommits[0])
+            commits = commits[:start_old]
+            print(f"{len(commits)} new commits")
+        else:
+            exit("No new commits")
+
+    # Access the variables.json and references.json files associated with each commit and store in dictionary
+    pool = multiprocessing.Pool(processes=multiprocessing.cpu_count())
```
agitter, line 57: I haven't used pools in a long time.  I see some examples using `with` statements (https://stackoverflow.com/questions/45718546/with-clause-for-multiprocessing-in-python) but do not know whether that is necessary.  Perhaps it is more robust in case the commit processing has an unexpected error.

analyze-ms-stats/calc-manuscript-stats.py
```diff
@@ -0,0 +1,126 @@
...
+        ax.minorticks_off()
+        ax.grid(color="lightgray")
+
+    ax.figure.savefig(args.output_figure + '.png', dpi=300, bbox_inches = "tight")
+    ax.figure.savefig(args.output_figure + '.svg', bbox_inches = "tight")
+
+    print(f'Wrote {args.output_figure}.png and {args.output_figure}.svg')
+
+    # Write json output file
+    manuscript_stats = growthData.iloc[0].to_dict()
```
agitter, line 97: When saving to json, let's name these variables to be easier to use as template variables within the text.  I'm using https://github.com/greenelab/covid19-review/blob/external-resources/csse/csse-stats.json as an example.  In this case, what do you think about:
```
  "stats_author_count": "52",
  "stats_clean_date": "September 2, 2021",
  "stats_iso_date": "2021-09-02",
  "stats_references": "1588",
  "stats_word_count": "134821"
```

Or we could use `growth_stats` as the prefix instead of `stats`.

Then we could use those variables instead of the hard-coded values in
> As of April 30, 2021, there were 50 authors, 1,428 references, and 131,949 words in the documents that make up the project.

agitter (changes requested):
I reviewed the code, and it looks good to me overall.  I did not try to run it locally or test the caching behavior.

This pull request could focus on adding the code, figures, and statistics.  Then I could follow up later to add that to an automated workflow and move the image and json outputs to `external-resources`.  That could happen after the camera ready deadline.

> I changed it to just use the absolute word count instead of dividing the word count by 1000 (I'm sure we can get this working that way if we feel strongly that we want it, but this was easier).

That works for me.

analyze-ms-stats/calc-manuscript-stats.py
```diff
@@ -0,0 +1,126 @@
...
+    start_time = time.time()
+    print(f"Using {multiprocessing.cpu_count()} CPUs")
+
+    # Read in list of all commits on this branch
+    with open(args.commit_list, "r") as commitFile:
+        commits = [c.strip() for c in commitFile.read().splitlines()]
+
+    # If this analysis has been run before, load in the list of commits analyzed
+    # and only analyze new commits
+    # Assumes no commits will be added retrospectively (to take advantage of linearity)
```
miltondp, line 42: One way to remove this assumption could be to create a `set` of old and new commits (I understand these are commits' hashes). Then you can process the commits that are new only. Assumbing both `commits` and `oldCommits` are both lists of commits' hashes, it would be something like: `commits_to_be_processed = set(commits) - set(oldCommits)`.

analyze-ms-stats/calc-manuscript-stats.py
miltondp: Yes, I think that's the only advantage of using `with` (not sure what happens if any of the tasks fails and you didn't close the pool). Usually, I prefer to use `futures`, like in [this function](https://github.com/hakyimlab/phenomexcan/blob/38390ac21987f1e72835c42919c53abd1a35cb7e/scripts/000_spredixcan/utils/generate_jobs.py#L78), `write_phenotypes_jobs`. But if this is already working fine, no need to change.

analyze-ms-stats/calc-manuscript-stats.sh
```diff
@@ -0,0 +1,19 @@
+# Calculate growth statistics for manuscript based on Manubot files
```
miltondp, line 1: A minor suggestion, in case you want the script to stop if any command fails (instead of continuing).

```suggestion
#!/bin/bash
set -e

# Calculate growth statistics for manuscript based on Manubot files
```

miltondp (approved):
Looks good to me, Halie. I left some comments, and I agree with @agitter suggestions. Regarding the parallelization of the code, I think the code is fine if it works for you. I left some comments regarding that as well.

analyze-ms-stats/calc-manuscript-stats.py
```diff
@@ -0,0 +1,126 @@
...
+    clean_date = variables['manubot']['date']
+    num_authors = len(variables['manubot']['authors'])
+    word_count = variables['manubot']['manuscript_stats']['word_count']
+
+    # Access files and data in references.json associated with each commit
+    referencesCommand = "git show " + commit + ":./references.json"
+    references = json.loads(subprocess.getoutput(referencesCommand))
+    num_ref = len(references)
+
+    return([date, clean_date, num_authors, word_count, num_ref])
```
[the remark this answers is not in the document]
rando2: Good call!

rando2:
@agitter I believe I'm implemented all the feedback from you and Milton! The only thing is that I need to switch the branch so that the commits are on top of the `external-resources` branch -- but doing that will close this PR and open a new one, so I wanted to make a note here!

analyze-ms-stats/environment.yml
```diff
@@ -0,0 +1,55 @@
+name: ms-stats
```
agitter, line 1: I don't think we'll need a new environment for this analysis, but let's defer that decision for when we integrate this with a scheduled workflow.  It's okay to commit this temporarily as a record of what packages are known to be compatible with the code.

agitter (commented):
Thanks for the updates @rando2.  I have a few more small comments.  It's okay with me that you'll create a new pull request when it's time to merge this with `external-resources`.

analyze-ms-stats/calc-manuscript-stats.py
```diff
@@ -0,0 +1,132 @@
...
+    return ({"stats_date": date,
+             "stats_clean_date": clean_date,
+             "stats_num_authors": num_authors,
+             "stats_num_words": word_count,
+             "stats_num_references": num_ref})
+
+def main(args):
+    '''Extract statistics from the output branch log'''
+
+    print("Using {0} CPUs".format(multiprocessing.cpu_count()))
```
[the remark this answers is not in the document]
rando2: It was working until I moved things into conda! I'm sure there's some package issue or something but it seemed easier to just swap it!

analyze-ms-stats/calc-manuscript-stats.py
```diff
@@ -0,0 +1,132 @@
...
+        exit(commit + " not found")
+
+    date = variables['pandoc']['date-meta']
+    clean_date = variables['manubot']['date']
+    num_authors = len(variables['manubot']['authors'])
+    word_count = variables['manubot']['manuscript_stats']['word_count']
+
+    # Access files and data in references.json associated with each commit
+    referencesCommand = "git show " + commit + ":./references.json"
+    references = json.loads(subprocess.getoutput(referencesCommand))
```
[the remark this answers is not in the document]
rando2: Thank you for catching this! I also moved the string creation outside of the try/except block, since that shouldn't fail (and definitely not with a JSON error).

analyze-ms-stats/calc-manuscript-stats.py
```diff
@@ -0,0 +1,132 @@
...
+    num_ref = len(references)
+
+    return ({"stats_date": date,
```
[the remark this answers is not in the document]
rando2: This data also gets converted into a pandas dataframe, so a dictionary seemed more intuitive to me! I should have specified that.

analyze-ms-stats/environment.yml
rando2: That sounds good to me!

agitter (approved):

rando2:
See #1034

PR #1034, opened 2021-09-13
#1019 for plotting project growth, but on external-resources branch

rando2:

@mprobson co-authored this PR
We tried to move the `ms-stats` branch over to `external-resources` and for some reason it would not let us rebase (other than manually going through each of 4011 commits 😬)
So instead we just moved the code itself over, effectively collapsing it into a single commit.
The revision history will be preserved in #1019 though!

### Related issues

#1019

### Checklist


agitter (approved):
I'm approving based on my review of the previous branch.

After submitting the DISCO manuscript, I'll work on rerunning this analysis in a workflow and using versioned figures.

PR #1035, opened 2021-09-13
Edits to master branch required for integration of #1034

rando2:

I believe this will add the manuscript growth statistics/figs (#1034) to the build

### Checklist


rando2:
This is failing but presumably there's no way it would succeed until #1034 merges. So I'm going to mark as draft until we can confirm what happens in the build.

agitter:
I merged #1034 and reran the GitHub Actions workflow here.  We'll have to hard-code the URL for the figures in the manuscript for now.  I'll update that later once the figures are periodically updated in a workflow.

agitter (approved):
The build passes, but that didn't test much.  Let's merge so we can see whether the nightly external resources workflow runs correctly.

PR #1036, opened 2021-09-13
collapse ref section by def, load scite badges on intersect

### LINKED ISSUE #1015: HTML load time
The HTML for the complete manuscript https://greenelab.github.io/covid19-review/ has gotten quite slow to load.  This came up as an issue during manuscript review (#1014).

The Chrome performance tool shows it takes about 40s to load on my machine
image

@vincerubinetti do you have any ideas about how to optimize load times?  Would it help if we collapse the appendix?  Disable scite  by default?  The javascript elements, like showing citation information on hover, can also be unresponsive in this huge manuscript.

vincerubinetti:
Closes #1015 

- allows the accordion plugin to accept default collapsed state for each heading \*
- only adds scite badge when citation comes into view
- make attributes plugin run on `DOMContentLoaded` (after page text and dom parsed but before scripts/styles/images/external resources load) instead of `load` (after everything is finished loading), to make sure it runs before other plugins \*
- cleans up a few lines (left-in console.logs, script type="module", etc) \*

\* changes we probably want to make upstream in rootstock


vincerubinetti:
@agitter it'd be helpful if you could clone this locally and build it and see if it works.

content/90.back-matter.md
```diff
@@ -1,5 +1,5 @@
 {% if format is defined and format != "tex"-%}
```
agitter, line 1:
```suggestion
{% if format is not defined or format is defined and format != "tex"-%}
```

I broke this when working on the LaTeX export.  It should be modified like this.

agitter (commented):
This is what I see in my local build:
image

Reference 2 should have a scite badge.  When should it appear?

I forgot that I made the References an h1 header for this manuscript.  When I modify the HTML output to make it an h2, the arrow looks collapsed but the content is shown.  The appendices are also h1 headers, which we may also want to collapse.

vincerubinetti:
> Reference 2 should have a scite badge. When should it appear?

This was a bug that should be fixed with my latest commit.

The accordion plugin wasn't written to support collapsing h1's. They weren't really meant to be used like this. It would help if the h2's (being generated?) under the appendixes could be collapsed instead.

Also I don't know why the references wouldn't be collapsed in your screenshot. The accordion plugin basically looks for the next h2 or h1 after the h2 you clicked, and hides everything in between. Is there some how another h1 or h2 being added in between ## References and the citations? Is the ## References in a container or something.

vincerubinetti:
Latest app veyor build shows the references section, written as a `## heading 2`, is showing up as an `<h1>` somehow... are you guys doing anything odd with the pandoc processing or something else that I'm not aware of.

image

And I had to manually show that because it's being hidden by the fact that it's `data-collapsed="true"` (the only elements not hidden by that are `<h2>`s).

agitter:
The heading level issue comes from this combination of pandoc settings
https://github.com/greenelab/covid19-review/blob/5611bd91a215753f643fadfa483ab07c3895f314/build/pandoc/defaults/common.yaml#L15-L19

Because we're numbering sections and splitting parts of this manuscript into individual DOCX sometimes, we adjusted the heading levels.

vincerubinetti:
> Because we're numbering sections and splitting parts of this manuscript into individual DOCX sometimes, we adjusted the heading levels.

So can we change ## references to a heading 3 so it can be collapsible. And are you able to add `<!-- $data-collapsed="true" -->` to all the headings under the appendices. Or do I need to make it possible for the accordion plugin to collapse heading 1's (significant work and testing).

agitter:
> And are you able to add `<!-- $data-collapsed="true" -->` to all the headings under the appendices that you want to hide.

Yes, I think we can do this with a regex.  We have a script that downloads these references from an external repository, and we're already doing some light processing on the content before committing the markdown here.  This sounds like the best solution.

> Or do I need to make it possible for the accordion plugin to collapse heading 1's (significant work and testing).

Let's not modify h1s.  That's too much work.

What do you see in the latest artifact `manuscript-1.0.4491-51f505b.html`?  I now have References as a h2 heading, but it isn't actually collapsed initially.  I also see whitespace for the scite badges, but the badges don't appear:
image

vincerubinetti:
Okay I fixed a few things.

Turns out the Scite script only scans the page after load, not continuously, so I had manually rescan for badges each time a new one comes into view. The badges should load okay now.

Also, the References section wasn't being collapsed properly because the attributes plugin that adds the override was being run after the accordion plugin that does the collapsing. That should be fixed.

I also added `data-collapsed="true"` directly to `<div id="refs">` and bumped the associated hiding style, a trick which should hopefully make it invisible immediately (see the comments).

vincerubinetti:
Note, we're waiting on a upstream Scite Badge change that would let us remove these lines: https://github.com/greenelab/covid19-review/pull/1036/files#diff-68837f720ce30c8aa59e5fb4fac66c3c37233da19e71d15686bb6de328984bd4R58-R61
https://github.com/greenelab/covid19-review/pull/1036/files#diff-68837f720ce30c8aa59e5fb4fac66c3c37233da19e71d15686bb6de328984bd4R71

content/90.back-matter.md
```diff
@@ -1,8 +1,19 @@
-{% if format is defined and format != "tex"-%}
-## References {.page_break_before}
+{% if format is not defined or format is defined and format != "tex"-%}
+### References <!-- $data-collapsed="true" --> {.page_break_before}
 
+<!-- 
+  This style exists in the accordion plugin already, but it will be loaded after
```
agitter, line 5: Is this a general issue with where the plugins are loaded, or is it a special case related to what we want to do with the collapsed section here?

agitter (approved):
This version looks good to me.  The references section is initially collapsed:
image

The badges appear:
image

The HTML artifact now loads in less than 10s in my browser.  @rando2 if this References modification looks good to you as well, we can merge this.  We can follow up in a new pull request to collapse the preprint reviews in the appendix.

content/90.back-matter.md
vincerubinetti: In general, I've tried to make the plugins such that they can work anywhere in the document, in case a manubot user changes that. This is a very specific instance of performance optimization; the `<style>` and `<div id="refs" data-collapsed="true">` would be done by the accordion plugin anyway because of the `### References <!-- $data-collapsed="true" -->`, they'd just be done a little bit further down the document.

vincerubinetti:
> we can merge this

Wait for my above comment regarding upstream changes. This will make the badges look nicer too.

vincerubinetti:
Thanks to Milo Mordaunt on the Scite team for quickly implementing our requested change that `insertBadges` not spending time redoing badges that have already been filled in.

This PR is ready for merge then. Though also, it'd be nice to know whether other people besides @agitter see performance improvements.

rando2:
I am actually having a lot of trouble loading the scite badges now -- in Firefox they are empty (0/0/0/0) or else `print: false screen: true` and in Safari, I've only been able to get them to render as `print: false screen: true`. The load time is still basically unusable in Safari and still very slow in Firefox :(

vincerubinetti:
@rando2 I have a feeling I know why this is. Here is an explanation for posterity or for those curious. Long story short, we should probably merge this as soon as possible.

The Scite people implemented a feature at my request, where calling `insertBadges` doesn't touch badges that have already been "filled in". I wanted this so that I could call `insertBadges` each time a new badge came into view, but without wasting time redoing (re-filling-in) all the badges that already came into view. They implemented this feature... but I bet the way they implemented it is by simply checking if the badge has any `innerHTML` content. They instruct you to put a blank `<div class="scite-badge" data-doi="blah"></div>` where you want the badge to appear, and once Scite adds a badge there, that `<div>` gets filled with a bunch of stuff to actually show the numbers/icons/etc. I was adding that `print: false screen: true` stuff to the `innerHTML` of the divs as a stupid debug message that I forgot to remove, so the Scite `insertBadges` probably skipped over them.

agitter:
> So they made an upstream change which breaks our current master branch, but this PR accounts for that and fixes it.

That means that it not only breaks the master branch of this review but also any other Manubot manuscripts using the scite plugin, right?  Do we need a way to fix other Manubot manuscripts using the plugin?

The latest artifact `manuscript-1.0.4506-5a7cb27.html` still works well for me in Chrome.

vincerubinetti:
> Do we need a way to fix other Manubot manuscripts using the plugin?

I'm not sure. Could you do me a favor (this is more difficult for me because I'm on a windows machine), and build the latest rootstock with Scite enabled, and see if the badges are stuck on `Loading Scite badge...` (the `innerHTML` I set). It might not be a problem because I'm not calling `insertBadges` there, the script is just scanning after page load. But I'm worried that under the hood Scite is just calling the same `insertBadges` function itself on page load.

bananaoomarang:
Hi just seeing this and chiming in as a developer from scite.

The change should not have broken existing integrations unless they were expecting `insertBadges` to reload all badges on the page. To my knowledge it should not have broken anything here as we do not look at the innerHTML to check if it's empty, the check is if `data-fetched` appears on the div (which we add after we have fetched the data once), not for innerHTML so it should not matter if the div is empty to begin with, or at least there should be no regression in this case.

vincerubinetti:
Apologies @bananaoomarang I should've checked with you before speculating.

Further, @rando2 I'm looking at master now and seeing that it is still working, at least on my machine, it just takes a really long time to load the badges. The `print: false` debug message still shows for a while, followed by a badge with 0's, and finally with real numbers.

Regardless of all this, this PR seems to be working and we should merge, @agitter

agitter:
I double-checked that the original scite plugin is still working.  My demo manuscript has a pull request (https://github.com/agitter/my-manuscript/pull/20) that is based on the latest rootstock.  You can check that the badges work in its [AppVeyor artifact](https://ci.appveyor.com/api/buildjobs/2nnj61lpbm48o56s/artifacts/manuscript-1.0.177-f2b8c2e.html).

I'm ready to merge but will wait to let @rando2 do it.  @rando2 you should be able to see the new badge loading behavior in this [HTML artifact](https://ci.appveyor.com/api/buildjobs/12cbf9ka8vf7lb97/artifacts/manuscript-1.0.4506-5a7cb27.html).

rando2:
Thank both! That HTML artifact is working great for me in Firefox and in Safari now! (I totally forgot it is in the zipped build, apologies, yesterday was a long day!)

PR #1041, opened 2021-09-15
Arrange CORD-19 figure panels vertically

agitter:
I rearranged the CORD-19 figure to better fit in the two-column format for the DISCO submission.  I made some modifications to regenerate figures locally that I need to revert before merging.  I also need to update the figure URL in the JSON file.


agitter:
@rando2 this change to the CORD-19 growth figure style is ready for review.  This is the version I used for the DISCO submission.

rando2 (approved):
Looks great! The thicker line makes a big difference, I like it a lot better this way!

PR #1042, opened 2021-09-16
Add therapeutics to latex build

rando2:

mSystems has changed their formatting and it's been really hard to figure out everything that needs to be changed, so instead I'm going to migrate the text into their latex template. Unfortunately it's in a different format than the CEUR one (like, nothing is a variable and I'm just not quite sure I know enough to do this correctly), so for now I'm going to hold off building this into the pipline -- just going to copy and paste the manuscript text for now and do the formatting manually.

### Checklist

### DIFF
--- a/content/individual-latex-manuscripts.txt
+++ b/content/individual-latex-manuscripts.txt
@@ -1,1 +1,2 @@
 methods
+pharmaceuticals


agitter (approved):
Oh no, that's too bad.  In our tex templates with the variables, @mprobson had to modify the original templates to add in the variable support.  If you want to go that route, I may be able to offer advice.  It may be worth the time if you are going to have to do this for every new mSystems submission.

Is there any way to work with the docx or is it hopelessly different?  I'm expecting the LaTeX conversion will be slow and time consuming with each submission.

If you are forced to go the LaTeX route, watch the URL citations carefully.  Those have not converted to .bib well in the DISCO submissions.  I also found it useful to add the versioned figure URL as a comment above each figure in the .tex file so that we know where it came from.

rando2:
Thanks @agitter, that makes sense! 🤦‍♀️ The documentation is tricky to navigate, but I think I've figured out what's going on with mSystems -- [they are format neutral for the initial submission, but have formatting requirements for resubmission](https://journals.asm.org/format-neutral-submissions). I think you're right that the URLs will be quite an endeavor in LaTeX. I was worried because our references are quite in the format they want, but they're pretty close, so hopefully this will be OK.

agitter:
Would it help if we try to modify the reference DOCX file that Pandoc uses when converting to DOCX?  If you're referring to the reference style in the DOCX, we can change that pretty easily with a new CSL file.

If it's the LaTeX references, we'll have to watch the CSL JSON to .bib conversion.  I'm guessing their LaTeX template specifies a citation style that will mostly work as long as our .bib file is complete.

rando2:
> Would it help if we try to modify the reference DOCX file that Pandoc uses when converting to DOCX? If you're referring to the reference style in the DOCX, we can change that pretty easily with a new CSL file.
> 
> If it's the LaTeX references, we'll have to watch the CSL JSON to .bib conversion. I'm guessing their LaTeX template specifies a citation style that will mostly work as long as our .bib file is complete.

That's great to know it should be feasible! Casey said to go with the whack-a-mole approach of only worrying about things as they complain about them, so fingers crossed they accept this version. If it does get bounced, I might need to look into making some edits to how things build, but for now hoping for the best! Have a great weekend Tony!

rando2:
Got it through without needing this! Whoo-hoo 🎉

PR #1062, opened 2021-10-27
Update 10.diagnostics.md

aimundo:

Added information about differences between RT-PCR, and qPCR, reorganized the flow of some paragraphs, and added information about infectious status of patients measured via qPCR.

### Checklist


agitter:
@aimundo did we ever get you added to the author list?  I don't see you listed in `content/metadata.yaml`. Let's follow up with a new pull request to include you there.

aimundo:
> @aimundo did we ever get you added to the author list? I don't see you listed in `content/metadata.yaml`. Let's follow up with a new pull request to include you there.

How weird, I thought I had added my info. Do I still need the pr to add it?

PR #1065, opened 2021-12-02
Update SMB affiliation

rando2:

@SiminaB I just wanted to confirm that this looks right! 

We can add it back in for future manuscript, but if it's OK to remove it for now, it will make generating correct versions a lot easier!

### Checklist

### DIFF
--- a/content/metadata.yaml
+++ b/content/metadata.yaml
@@ -191,7 +191,6 @@ authors:
       confirmed: Yes
     affiliations:
       - Innovation Center for Biomedical Informatics, Georgetown University Medical Center, Washington, District of Columbia, United States of America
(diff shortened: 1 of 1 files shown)


SiminaB (commented):
Looks good! Can we keep it as is for therapeutics (so have Georgetown + AZ) but remove it for pathogenesis (Georgetown only, AZ just as COI)? If we're not planning to update therapeutics very soon though, it should be OK to temporarily remove it from Github, then add it back later.

rando2:
@SiminaB Sounds good to me! I will add it back in after I generate a new PDF for arxiv. Would you mind approving this PR? I'll tag you again when I put it back!

SiminaB (approved):

PR #1069, opened 2021-12-13
Visualize vaccine distrubtion by technology

rando2:

@RLordan I think we had discussed this ages ago, but here are some figures to show the distribution of each vaccine type. I was thinking to add them to the manuscript where we end up talking what has been approved. The colors correspond to how many different vaccines of each type have been rolled out in each country (according to OWID)

RNA
inactivated_whole-virus
non-replicating_viral_vector
protein_subunit

This PR (especially the code) needs to be cleaned up, but I wanted to get your thoughts on the figures first. Originally I was trying to make a four-panel graph, but that is hard to make forwards-compatible (by which I mean, if more types of vaccines are approved, we would need more panels).

I'll tag someone for technical feedback after Ronan approves the actual objective of the PR, which was to visualize some stats he collected!

### Checklist


agitter:
Very cool!  What are the units on the colorbars?

RLordan (approved):
Hey @rando2 exactly the type of graph I was hoping for. Although I have the same question as @agitter about the units. I am sure this can be "standardized" across the four maybe?

rando2:
@RLordan @agitter So it's the number of vaccines of that type available. This may or may not be the most interesting thing to represent -- I think ideally I'd want to show each vaccine candidate and where it's available (e.g., color in Pfizer blue and Moderna red and countries with both purple). 

At a minimum, I'm thinking I should make the scale discrete instead of continuous so that we can standardize them -- I have had a lot of trouble in the past getting colors on scales to line up across subplots!

agitter:
> it's the number of vaccines of that type available. 

That makes sense

>  make the scale discrete instead of continuous

That's a great idea to help make the different panels more directly comparable.

RLordan:
> @RLordan @agitter So it's the number of vaccines of that type available. This may or may not be the most interesting thing to represent -- I think ideally I'd want to show each vaccine candidate and where it's available (e.g., color in Pfizer blue and Moderna red and countries with both purple). 
> 
> At a minimum, I'm thinking I should make the scale discrete instead of continuous so that we can standardize them -- I have had a lot of trouble in the past getting colors on scales to line up across subplots!

Ahhh! makes sense. That idea would be cool.

rando2:
@RLordan @agitter update here! Let me know what you think, I still need to do some clean-up before it's ready for review
inactivated_whole-virus
non-replicating_viral_vector
protein_subunit
RNA
!

agitter:
That style looks great to me, and your new proposed title is also clear.

RLordan:
This is great @rando2.

rando2:
Thank you both! I'll try to get this PR finalized ASAP (this week is a little weird, I'm out of the office today and tomorrow).

rando2:
This should be ready for review now!

RLordan:
> This should be ready for review now!

I think I approved already, but this is clearer now. I can't comment on code, but the choropleths look great

ajlee21 (approved):
This are nice plots! Very interesting to see the distribution of what is available across the country. I definitely agree with the title change (I initially thought this was the number of people vaccinated with each vaccine type)

owiddata/generate-owiddata-stats.py
```diff
@@ -48,22 +77,91 @@ def main(args):
...
+    for iso, vaccines in allVaxByCountry.items():
+        for vax in vaccines.split(","):
+            vax = vax.strip()
+            countryCodes = countryByVax.get(vax, [])
+            countryByVax[vax] = countryCodes + [iso]
+
+    # Add countries to vaccine platform info and plot each vaccine type
+    vaxPlatforms['countries'] = vaxPlatforms.index.map(countryByVax)
+
+    for platform in set(list(vaxPlatforms["Type"])):
```
agitter, line 124:
```suggestion
    for platform in set(vaxPlatforms["Type"]):
```
Not a big deal, but is the `list()` needed here?

owiddata/generate-owiddata-stats.sh
```diff
@@ -3,18 +3,20 @@
...
 
-
 # Get the most recent commit affecting this file and store the commit sha and date
 # See https://stackoverflow.com/questions/50194241/get-when-the-file-was-last-updated-from-a-github-repository/50204589
 OWID_COMMIT_JSON=$(curl -sS "https://api.github.com/repos/owid/covid-19-data/commits?path=$OWID_REPO_PATH&per_page=1")
 export OWID_COMMIT_SHA=$(echo $OWID_COMMIT_JSON | python -c "import sys, json; print(json.load(sys.stdin)[0]['sha'])")
 export OWID_COMMIT_DATE=$(echo $OWID_COMMIT_JSON | python -c "import sys, json; print(json.load(sys.stdin)[0]['commit']['author']['date'])")
 
+# Input file(s)
+VACCINE_PLATFORMS=owiddata/vaccine_platforms.csv
```
agitter, line 13: Where does this csv come from?  Is it manually created once?  Will it ever need to be updated?

agitter (commented):
These maps are really great.  The code changes look good to me overall.  I did have one question about the input csv file and whether we'll need to maintain that.

owiddata/generate-owiddata-stats.py
rando2: Nope, it's not needed! For some reason I did that while debugging. I'll push a new commit to remove it.

owiddata/generate-owiddata-stats.sh
rando2: We will need to update it :( I haven't found a good external resource to unify these. When new vaccines are added in OWID, this will exit with an error saying to update the CSV. (Line 109 in the python script)

rando2 (commented):
Thank you so much @agitter!

owiddata/generate-owiddata-stats.sh
agitter: That approach seems good enough for now.  We may not notice if these vaccine figures stop updating because of that error, but we can check periodically.

agitter (approved):

PR #1082, opened 2022-02-22
Add pharmaceuticals to docx build

rando2:

@Jeff-Field has asked for a word document of the pharmaceuticals text, so I'm adding it the build!

### Checklist

### DIFF
--- a/content/individual-docx-manuscripts.txt
+++ b/content/individual-docx-manuscripts.txt
@@ -1,2 +1,2 @@
-pathogenesis
 vaccines
+pharmaceuticals


rando2:
Looks like there is an incompatibility between the new version of jinja2 and a utility called "soft\_unicode" that was preventing this from building, see \<https://github.com/aws/aws-sam-cli/issues/3661\>

RLordan (approved):

agitter (approved):
That new error is annoying.  I should update to the latest version of rootstock, which isn't affected per https://github.com/manubot/rootstock/issues/459.

Pinning the `markupsafe` version is a good solution in the meantime.

PR #1083, opened 2022-03-07
Update protein subunit

rando2:
### Description of the proposed additions or changes

Hey @RLordan, here are some edits to the protein subunit section. I am going through again (😬) to try to trim and move text from the paper as a whole, so there will probably be more PRs this week with those edits! I had a grant due 2/28 but I'm back to aggressively trying to wrap this manuscript up now!

### Checklist

- [ ] Text is formatted so that each sentence is on its own line.

- [ ] Pre-prints cited in this pull request have a GitHub issue opened so that they can be reviewed.

PR #1087, opened 2022-03-08
Update rootstock on 2022-03-08

agitter:
I noticed a huge number of DOI citation warnings in recent builds.  These are due to changes in the external services that Manubot uses for DOI metadata.  Updating Manubot should reduce these warnings.

We'll also gain the ability to build the manuscript on demand from https://github.com/greenelab/covid19-review/actions without creating a commit or pull request.

This updates the build script so we should carefully check the build log and outputs before merging.


rando2:
@agitter the builds look mostly OK (the pdf is justified weirdly and I can't remember if it always looked like that) but there are a lot of errors. It seems like the incorrect DOIs are returning the full HTML of the DOI Not Found page on doi.org and there are also lots of RawBlock errors. I don't feel confident about whether this is expected or if something it going awry!

agitter:
I compared the AppVeyor PDF here with the [current PDF](https://greenelab.github.io/covid19-review/manuscript.pdf) and the justification looks the same.  It does seem weird, but that isn't due to this pull request.

The substantially more verbose output for DOI errors is annoying.  The primary DOI metadata provider changed, so it could be due to a different response when an invalid DOI is provided.  The two errors for `doi:10.1093/cid/ciab229/6167855` and `doi:10.3389/fimmu.2020` aren't recognized DOIs at doi.org, so we'll need to fix those separately.

Overall the build and artefacts look okay to me.  I think we can proceed to merge, and tag me if we notice any build issues going forward.

rando2 (approved):

rando2:
> The two errors for `doi:10.1093/cid/ciab229/6167855` and `doi:10.3389/fimmu.2020` aren't recognized DOIs at doi.org, so we'll need to fix those separately.

These are now fixed in #1083

PR #1091, opened 2022-03-14
Edits to Diagnostics ms

rando2:
### Description of the proposed additions or changes

This PR seeks to flesh out all of the sections of the diagnostics manuscript, pending the additions planned by @aimundo and @RLordan 

Feedback is very welcome!

### Checklist

- [ ] Text is formatted so that each sentence is on its own line.

- [ ] Pre-prints cited in this pull request have a GitHub issue opened so that they can be reviewed.

PR #1092, opened 2022-03-15
Reorganize sections within diagnostics

rando2:
The edits in #1091 depend on the sections being moved, so it would be preferable to move them first and then edit them!

### Checklist


rando2:
@RLordan it's just rearranging the sections (no edits!) to make it easier once we start reviewing the substantive changes in #1091. So if you don't mind reviewing my rearranging, that would be amazing!!!

RLordan (approved):

PR #1104, opened 2022-03-23
Update to processing of external data about vaccines

### LINKED ISSUE #1101: External resources workflow failing due to missing vaccine platforms
Our external resources GitHub Actions workflow has been failing for about 2 months.
- [Last good run](https://github.com/greenelab/covid19-review/runs/4829438178?check_suite_focus=true)
- [First bad run](https://github.com/greenelab/covid19-review/runs/4834925613?check_suite_focus=true)

It happens in the "Update OWID data" step:
```
Generating Our World in Data COVID-1[9](https://github.com/greenelab/covid19-review/runs/4834925613?check_suite_focus=true#step:7:9) vaccine statistics
Missing platform information for Novavax
Error: Process completed with exit code 1.
```

Based on this comment, it looks like `vaccine_platforms.csv` needs to be manually updated.  Is that correct @rando2?
https://github.com/greenelab/covid19-review/blob/b493cdf544a0ea04d3feb53a908a21790fcb457b/owiddata/generate-owiddata-stats.py#L106-L109

In the latest runs, there are multiple missing platforms.

rando2:

This PR updates the OWID external analysis to include data about each vaccine scraped from https://covid19.trackvaccines.org/
This will allow us to avoid needing to manually update the types of each vaccine
Additionally, I added in the code needed to store the URL associated with each vaccine in case we want to pull other data from that site down the road. (Since most of the data is on a separate page where the url contains an arbitrary number assigned to each vaccine)
I also added some new outputs to the json file, including markdown tables listing approved vaccines in each category

The map file names for the inactivated whole virus vaccines has changed. The incorrect reference is removed in #1105

One other concern: I manually added the packages I used to environment.yml. Checking the environment in conda generates a huge list because it's the project as a whole. I think it will be clear whether this worked based on whether it works OK, but I would guess there's likely a better way to do this!

### Related issues

closes #1101 

### Checklist


owiddata/generate-owiddata-stats.py
```diff
@@ -45,6 +73,52 @@ def convert_date(git_date):
...
+    cards = body.find_all('li')
+
+    # Iterate through the cards to extract information
+    # Tags were identified empirically and are not self-evident
+    for card in cards: # find all element of tag
+        if card.find('a', {"class": "icon-link"}) is not None:
+            vaccine_type = card.find('a', {"class": "icon-link"}).get_text()
+            if vaccine_type.upper() != vaccine_type: #DNA, RNA, VLP
+                vaccine_type = vaccine_type.lower()
+            vaccine_category = platforms(vaccine_type)
```
agitter, line 94: What's the difference between a type and a category?

owiddata/generate-owiddata-stats.py
```diff
@@ -45,6 +73,52 @@ def convert_date(git_date):
...
+    return vaccine_df
+
+def create_table(vaccine_df, category):
+    """For each vaccine type, select a subset of the vaccine information table
+    Input: dataframe, string
+    Returns: string representing a table in markdown"""
+    vaccines = vaccine_df[vaccine_df["Category"] == category]
+    numTypes = len(set(vaccines["Type"].to_list()))
+    if numTypes > 1:
+        return vaccines[["Company", "Type"]].to_markdown()
```
agitter, line 118: I'm not sure how the JSON will store the markdown.  We can see what happens and adjust if needed.

owiddata/generate-owiddata-stats.sh
```diff
@@ -9,9 +9,8 @@ OWID_COMMIT_JSON=$(curl -sS "https://api.github.com/repos/owid/covid-19-data/com
 export OWID_COMMIT_SHA=$(echo $OWID_COMMIT_JSON | python -c "import sys, json; print(json.load(sys.stdin)[0]['sha'])")
 export OWID_COMMIT_DATE=$(echo $OWID_COMMIT_JSON | python -c "import sys, json; print(json.load(sys.stdin)[0]['commit']['author']['date'])")
 
-# Input file(s)
+# The output filenames
 VACCINE_PLATFORMS=owiddata/vaccine_platforms.csv
```
agitter, line 13: Do you want to commit this output file?  We'll have to modify the workflow to add it.

agitter (commented):
This looks good to me overall.  I didn't try running the code.  You could add the .json file to the commit so we can inspect it before merging.

Are you concerned the that the conda environment may not build now that you added new packages?  We could wait to see what happens in GitHub Actions or run a local dry run to check whether the versions can all be resolved.

owiddata/generate-owiddata-stats.sh
rando2: I'll update to delete it in the .sh file! It's not needed anymore

owiddata/generate-owiddata-stats.py
rando2: I changed them to "platforms" and "platform types" to try to clarify this, since I'm not totally sure on the correct terminology. So for example, plasmid vectored, replicating DNA, and non-replicating DNA would all be types of DNA vaccines

owiddata/generate-owiddata-stats.py
rando2: I'm curious if this will work!!! In theory I think the JSON looked ok but there are some special characters that could make things interesting

owiddata/generate-owiddata-stats.py
```diff
@@ -4,11 +4,39 @@
...
+             "non replicating viral vector": "DNA",
+             "replicating viral vector": "DNA",
+             "inactivated": "whole virus",
+             "live attenuated": "whole virus"
+             }
+
+    # If they add a new category (which seems unlikely), handle & throw error
+    if vaxtype not in types.keys():
+        print("Unknown vaccine platform:", vaxtype)
+        return vaxtype
```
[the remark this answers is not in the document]
rando2: I don't think this should ever happen because I think it handles all the major categories of vaccines that currently exist? I just wrote it in as a just-in-case. So I think either would be fine!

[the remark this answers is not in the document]
agitter: In that case, I prefer we exit with an error.  That should happen rarely if ever, so we would want to know about the new major vaccine category.

agitter:
This is looking good.  Once the code is stable, could you please run it locally and commit the output images and json file so we can inspect them before merging?

owiddata/owiddata-stats.json
```diff
@@ -1,22 +1,33 @@
...
-  "owid_inactivated_whole_virus_map": "https://github.com/greenelab/covid19-review/raw/ee30031393799c211a140f9be7eefbed12ef0e1b/owiddata/maps/inactivated_whole_virus.png",
-  "owid_most_recent_date": "March 27, 2022",
-  "owid_non_replicating_viral_vector_count": 5,
-  "owid_non_replicating_viral_vector_countries": 200,
-  "owid_non_replicating_viral_vector_map": "https://github.com/greenelab/covid19-review/raw/ee30031393799c211a140f9be7eefbed12ef0e1b/owiddata/maps/non_replicating_viral_vector.png",
-  "owid_protein_subunit_count": 9,
-  "owid_protein_subunit_countries": 21,
-  "owid_protein_subunit_map": "https://github.com/greenelab/covid19-review/raw/ee30031393799c211a140f9be7eefbed12ef0e1b/owiddata/maps/protein_subunit.png",
+  "owid_RNA_count": 3,
+  "owid_RNA_countries": 0,
```
agitter, line 3: These country counts are lower than they were before, and sometimes 0.  Do you know what changed?

environment.yml
```diff
@@ -8,9 +8,13 @@ dependencies:
...
+  - geopandas=0.10.2
   - seaborn=0.11.1
+  - tabulate=0.8.9
+  - urllib3=1.26.4
   - pycountry==20.7.3
   - pip:
     - git+https://github.com/manubot/manubot@a57ccf0be6972329ff3010eaaa0c5df7ccebb2d5
     - wget==3.2
+    - bs4==0.0.1
+    - html5lib
```
agitter, line 20: Could you please pin the `html5lib` version?

owiddata/owiddata-stats.json
rando2: Ahhh! It's a merge issue for the two data sources. I'll add in another function to manually align their nomenclature.

agitter (approved):
Thanks.  I think this looks ready to merge.

rando2:
@agitter I think this aligns the datasets well now. I did write in a fairly bulky output so that we can investigate things that seem weird (above). It prints all of the info for vaccines in the VIPER dataset that aren't found in the OWID data. Logically that is the only direction we need (since VIPER tracks candidates and OWID only tracks shots in arms) but I'm also happy to add in another check if we'd feel better checking both directions!

agitter:
I trust your judgement about how to best confirm the matching across data sources worked well.  This seems good enough to me.  The country counts in the json output have been restored to their expected numbers.

agitter:
I manually triggered the external resources build from [here](https://github.com/greenelab/covid19-review/actions/workflows/update-external-resources.yaml) so that the figure URLs in the json file get updated with actual commits.  That will also let us confirm everything runs in the updated environment.

agitter:
The workflow ran successfully, but it took 5 hours (!) for the conda environment to install.  I'm not sure what would cause it to be so slow.

We could switch to mamba at some point.

rando2:
@agitter Oh my gosh, five hours??? When I build my conda environment.yml file locally, there are a lot of other dependencies, so it does seem like a complex environment. Sorry if these changes are slowing things down!!!!

PR #1105, opened 2022-03-23
fix reference to figure

rando2:

Handles the error introduced to the vaccines manuscript by the changes in #1104

### Checklist


agitter (approved):
Looks good to me, but we should merge #1104 first.

PR #1142, opened 2022-05-03
Add references back to PDF output

### LINKED ISSUE #1133: References missing in PDF
I noticed in #1132 that the references do not appear in the PDF output.  The header prints but the content is missing
image

@vincerubinetti do you have any ideas what could be causing this?  The build is generating the PDF with the Athena Docker container.  We made some changes in #1036, but that isn't necessarily the source of the problem.

The current rootstock PDF is not affected.

agitter:

Attempts to add references back to the PDF output by removing the automatically collapsed references section

### Related issues

Closes #1133  
Implements suggestion from https://github.com/greenelab/covid19-review/issues/1133#issuecomment-1115036957

### Checklist

### DIFF
--- a/content/90.back-matter.md
+++ b/content/90.back-matter.md
@@ -1,19 +1,8 @@
 {% if format is not defined or format is defined and format != "tex"-%}
-### References <!-- $data-collapsed="true" --> {.page_break_before}
+### References {.page_break_before}
 
-<!-- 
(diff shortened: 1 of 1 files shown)


agitter:
@vincerubinetti your suggestion had the desired effect.  The PDF and HTML references both look good now.

PDF:
image

HTML:
image

PR #1152, opened 2022-06-04
Switch external resources workflow to mamba

agitter:
The conda environment for the external resources no longer builds in GitHub Actions.  See a [recent build failure](https://github.com/greenelab/covid19-review/runs/6734550792?check_suite_focus=true).

I'm following https://github.com/conda-incubator/setup-miniconda#example-6-mamba to try to use mamba instead.

We may have to merge, test, and repeat until this works.

### DIFF
--- a/.github/workflows/update-external-resources.yaml
+++ b/.github/workflows/update-external-resources.yaml
@@ -16,16 +16,24 @@ jobs:
         uses: actions/checkout@v2
         with:
           ref: external-resources
-      - name: Install environment
+      - name: Install environment # Use mamba as in https://github.com/conda-incubator/setup-miniconda#example-6-mamba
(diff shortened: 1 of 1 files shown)


agitter:
I tested this conda setup in https://github.com/agitter/my-manuscript/pull/23

It looks like it is working.  The build log shows
```
Updating 'manubot' env from conda env update...
  /usr/share/miniconda/condabin/mamba env update --name manubot --file build/environment.yml
```

After we merge this, we'll need to work on the [`UnsatisfiableError`](https://github.com/greenelab/covid19-review/actions/runs/2440288897) from the external resources environment.

rando2 (approved):
Hooray for the 3min build, even if it's just the rootstock!

PR #1153, opened 2022-06-06
Resolve conda environment errors

agitter:
This environment worked in https://github.com/agitter/my-manuscript/pull/23.  It has only minor changes to versions so I'm hoping it would affect the script behavior.  It is paired with the switch to mamba in #1152, which now builds the environment in 3 minutes.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -9,8 +9,8 @@ dependencies:
   - pip=20.0
   - python=3.7.6
   - geopandas=0.10.2
-  - shapely=1.8.2
-  - fiona=1.8.21
+  - shapely=1.7.1
+  - fiona=1.8.17
   - python-levenshtein=0.12.2
   - seaborn=0.11.1
   - tabulate=0.8.9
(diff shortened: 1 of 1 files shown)


rando2:
I have also been working on this today so I'll make sure everything runs!

rando2 (approved):
This builds! @agitter can I ask about your process in figuring out which packages to downgrade? Did you just get the first " none of the providers can be installed" error and then check which packages had dependencies on it? In retrospect that's what I would think would work best, but yesterday I didn't think of it (I was trying to figure out how to upgrade the un-upgradable packages).

agitter:
At first I started trying to change package versions one by one to based on those error messages.  However, I quickly gave up and removed all of the new packages, then added them back without versions to see if any combination of versions was compatible, and then pinned the new versions once the environment resolved.  You can see my trial and error process in the commits in https://github.com/agitter/my-manuscript/pull/23.

rando2:
Thank you so much @agitter!!! That's so helpful, I am still in the learning phase with conda!

PR #1170, opened 2022-08-09
Bug fix: worldwide vaccine stats from OWID

rando2:

There have been a lot of NaNs lately and they were breaking this code. This should handle those errors.

### Checklist


owiddata/owiddata-stats.json
```diff
@@ -5,29 +5,29 @@
...
   "owid_protein_subunit_count": "10",
   "owid_protein_subunit_countries": "41",
-  "owid_protein_subunit_map": "https://github.com/greenelab/covid19-review/raw/94d16532b2ab833c40757b7807851db351668eed/owiddata/maps/protein_subunit.png",
+  "owid_protein_subunit_map": "https://github.com/greenelab/covid19-review/raw/$FIGURE_COMMIT_SHA/owiddata/maps/protein_subunit.png",
   "owid_total_countries": "223",
-  "owid_total_vaccinations": "12.0 billion",
   "owid_vaccine_counts": "29",
+  "owid_worldwide_daily_rate": "808.0 per million",
+  "owid_worldwide_date": "August 2, 2022",
+  "owid_worldwide_total_vaccinations": "12.0 billion",
```
agitter, line 30: Will we need to check in the manuscript if this variable has a valid value or a message like "Cannot convert count N/A to billions"?  I'm wondering if this message is going to be read by maintainers, in which case it looks good, or by a downstream Jinja conditional statement, in which case we can simplify it.

agitter (approved):

PR #1180, opened 2022-10-06
Make vaccines maps easier to read

rando2:

These edits do the following:
1. Edit code to set color scale within each paper
2. Increase size of map and the dpi to see how it looks

### Checklist


agitter (approved):
I don't know `plydata` syntax, but the code updates look reasonable.  Could you please commit the updated figures so we can see the diff?  The new example figures you copied look good.

agitter:
Thanks, these do look nice.  The updated scale is helpful for the novel vaccines figures.

Do you think we need to make the title (and scale bar?) text a larger font now that the figure size is bigger overall?  Once these are resized in the manuscript, the font sizes might look small.

rando2:
@agitter good point! I also caught that "NaN" values are causing an issue in the json (that I don't think actually affects the manuscripts, but seems worth fixing) so I'll tag you when I have both of these sorted out!

RLordan (approved):
Super improvement, thanks @rando2

rando2:
Thank you @RLordan and @agitter! What do you think of this?

agitter:
That title size looks good to me

PR #1181, opened 2022-10-10
Add DBMI affiliation

rando2:

I noticed we don't have DBMI listed! I also added "Anschutz" to the other CU affiliations.

### Checklist


agitter (approved):
Do you want to sort it so the Penn affiliation isn't between the Colorado affiliations? It won't really matter either way.

rando2:
Thanks @agitter, I think it's fixed now!

cgreene (approved):
🙌

PR #1192, opened 2023-01-10
Small updates to legend

rando2:

I dug through the code thinking the scale on the choropleths was off, but it looks like it is totally reasonable -- we just have to note why it is the way it is. I also added the VIPER url a few extra places. I'm not sure how to note that it stopped updating December 2, 2022 -- might be OK since we are not too far out from that date.

### Checklist


cgreene (approved):

agitter:
>  I'm not sure how to note that it stopped updating December 2, 2022

I don't think you have to. It looks like you were careful to insert the date stamp of the VIPER data when we use VIPER data, so it will be evident as that date stops updating.

agitter (approved):

PR #1194, opened 2023-01-13
Update rootstock on 2023-01-13

agitter:
This update includes support for corresponding authors, pinned publication dates, and other rootstock features. However, it contained more merge conflicts than I expected. Let's not merge it until after the traditional vaccines manuscript is completely finished (i.e. after I submit a separate pull request later today) in case it introduces formatting problems. I want to closely inspect the artifacts and the diff.

I'm attempting to comment out the new support for a pinned publication date. That's not relevant here where we have multiple publications. I'm also leaving out the corresponding author text from the front matter.

We will also be able to schedule manuscript builds now, for example on a weekly basis, which may be nice for this project.


agitter:
This seems to be working now after upgrading to the latest version of pandas. HTML artifact from this pull request:
image

HTML artifact from #1195 (uses the `date` from the updated manubot package but not the updated rootstock):
image

We could use the new corresponding author features, but I don't think it's necessary.

rando2:
I just merged #1199 so I might hold off an merge this once we're confident that change hasn't broken anything!

rando2 (approved):

rando2:
Manuscript looks good so I'm going to merge! 🤞

PR #1199, opened 2023-01-14
Switch external resources workflow to mambaforge

### LINKED ISSUE #1198: External resources workflow broke on 2023-01-13
Last good run: https://github.com/greenelab/covid19-review/actions/runs/3897878268
First broken run: https://github.com/greenelab/covid19-review/actions/runs/3907150141

It's failing at the `Adding tools to 'base' env...` step before it even uses our environment file.

Possibly related to `https://github.com/conda-incubator/setup-miniconda/issues/274`.

It may be hard to debug this via pull request because it needs to run on GitHub Actions. I might have to make commits directly on the `external-resources` branch and try running the workflow to see what happens.

agitter:
First attempt to resolve #1198.


rando2:
Looks good aside from some warnings (`'type' is a required property )` that I don't think are related, and which I'm making a separate PR to address.

agitter:
I manually ran the external resources workflow, and this successfully fixed the issue: https://github.com/greenelab/covid19-review/actions/runs/3941944026/jobs/6744997298

The two `mamba config` calls are ignored because it turns out `mamba` does not support them. I'm going to leave them for now in case we switch back to `conda` later.

@rando2 let me know if you see anything unusual in the build log messages linked above or in the generated commits fd0a82d9d43eb94beebd85f37d27d1652324cf67 and 347620682d254790e3b7e861f1d9ef667b41adf3

PR #1201, opened 2023-01-17
Clean up warnings

rando2:

There are some warnings in the (now very fast!) manubot build. I'm trying to debug here but I think they're all caused by webpages that manubot couldn't read, so I added manually.

### Checklist

### DIFF
--- a/.github/workflows/update-external-resources.yaml
+++ b/.github/workflows/update-external-resources.yaml
@@ -16,23 +16,23 @@ jobs:
         uses: actions/checkout@v2
         with:
           ref: external-resources
-      - name: Install environment # Use mamba as in https://github.com/conda-incubator/setup-miniconda#example-6-mamba
+      - name: Install environment # Use mamba as in https://github.com/conda-incubator/setup-miniconda/issues/274 and https://github.com/conda-incubator/setup-miniconda#example-10-miniforge
         uses: conda-incubator/setup-miniconda@v2
         with:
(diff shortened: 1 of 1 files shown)


agitter:
```
Run bash build/build.sh
Traceback (most recent call last):
  File "<string>", line 1, in <module>
KeyError: 'commit'
Error: Process completed with exit code 1.
```
We had this build failure on another recent pull request. That time I reran the failed job, and it completed successfully. It's somewhat concerning if it is recurring sporadically.

rando2:
@agitter Interesting. I pulled the branch locally and it's running fine. Based on the fact that it's a key error, I would guess something is buggy with the way the branch is being read in by sys and not the build.sh script itself? I'll keep an eye on this and see if a pattern emerges.

agitter (approved):
The error also isn't affecting AppVeyor builds, so I'm planning to ignore it for now.

agitter:
I checked the script and it must be coming from this line:
```
EXTERNAL_RESOURCES_COMMIT=$(curl -sS https://api.github.com/repos/greenelab/covid19-review/branches/external-resources | python -c "import sys, json; print(json.load(sys.stdin)['commit']['sha'])")
```
If something sporadically fails with that call to the GitHub API, we won't be able to parse the commit. It's okay to fail in that case because we need access to the commit from the external resources branch. This could have to do with GitHub stability or load. We could make the error message more descriptive.

Issue #1203, opened 2023-01-23
"Commit" not recognized in build.sh


rando2:
We have been seeing Manubot fail frequently with a keyerror because of 'commit' here:

> \# Log the external-resources commit used when building the manuscript
EXTERNAL_RESOURCES_COMMIT=$(curl -sS https://api.github.com/repos/greenelab/covid19-review/branches/external-resources | python -c "import sys, json; print(json.load(sys.stdin)['commit']['sha'])")
echo >&2 "Using external-resources commit $EXTERNAL_RESOURCES_COMMIT"

I'm guessing this means that the curl command is returning something that either a) isn't a dictionary, or b) doesn't contain a key named `commit`. I tried to replicate on the command line and rand it about 20 times without it failing. I'm making this issue as a place to stash ideas for how to troubleshoot or fish this!

rando2:
![](https://user-images.githubusercontent.com/8975841/214147133-3822ce31-30be-43e2-be83-c0f0a377abb3.png)

agitter:
Maybe the next thing to do would be to make this line a tiny separate python script that we call so it can have some reasonable error handling and tell us what it gets back from curl when the response is unexpected.

rando2:
I will finalize this tomorrow -- much cleverer approach than piping a bunch, which is what I was originally doing!

PR #1204, opened 2023-01-24
Test build error with verbose error handling

rando2:

See #1203

### Checklist

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -12,7 +12,7 @@ export TZ=Etc/UTC
 export LC_ALL=en_US.UTF-8
 
 # Log the external-resources commit used when building the manuscript
-curl -sS https://api.github.com/repos/greenelab/covid19-review/branches/external-resources > commitinfo.json
+curl -sS --retry 3 --retry-max-time 30 https://api.github.com/repos/greenelab/covid19-review/branches/external-resources > commitinfo.json
 EXTERNAL_RESOURCES_COMMIT=$(python build/read-commit.py commitinfo.json)
 echo >&2 "Using external-resources commit $EXTERNAL_RESOURCES_COMMIT"
 
--- a/build/build.sh
+++ b/build/build.sh
@@ -12,7 +12,8 @@ export TZ=Etc/UTC
 export LC_ALL=en_US.UTF-8
 
 # Log the external-resources commit used when building the manuscript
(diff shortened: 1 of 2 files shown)


rando2:
@agitter hilariously, this now won't fail. I ran it three times and they all built successfully!

build/read-commit.py
```diff
@@ -0,0 +1,11 @@
+import sys
+import json
+import os
```
agitter, line 3:
```suggestion
```

build/read-commit.py
```diff
@@ -0,0 +1,11 @@
...
+
+with open('commitinfo.json', 'r') as infile:
```
agitter, line 5:
```suggestion
with open(sys.argv[1], 'r') as infile:
```
It looks like you wanted to pass the filename as an argument

build/build.sh
```diff
@@ -12,7 +12,8 @@ export TZ=Etc/UTC
 export LC_ALL=en_US.UTF-8
 
 # Log the external-resources commit used when building the manuscript
-EXTERNAL_RESOURCES_COMMIT=$(curl -sS https://api.github.com/repos/greenelab/covid19-review/branches/external-resources | python -c "import sys, json; print(json.load(sys.stdin)['commit']['sha'])")
+curl -sS https://api.github.com/repos/greenelab/covid19-review/branches/external-resources > commitinfo.json
```
agitter, line 15:
```suggestion
curl -sS --retry 3 --retry-max-time 30 https://api.github.com/repos/greenelab/covid19-review/branches/external-resources > commitinfo.json
```
Should we add a retry? I'm not sure what the problem is or if this will help.

build/read-commit.py
```diff
@@ -0,0 +1,11 @@
...
+    data = json.load(infile)
+    if 'commit' in data:
+        print(data['commit']['sha'])
+    else:
+        print(data)
+        exit(1)
```
agitter, line 11:
```suggestion
        sys.exit(f'Unexpected response from external-resources branch:\n {data}')
```
When I tested locally by changing the key in the if statement, the print statement didn't print anything because exit was called before it flushed. Is this better?

agitter (changes requested):
I wasn't able to actually trigger the problem locally either. At least this will give us a dump of the json contents if we get a curl response but it doesn't contain the commit information. If curl fails and we don't write `commitinfo.json`, we'll need a more general approach.

rando2:
Thank you @agitter! Let's see what happens now!

agitter (approved):
Let's merge and see what happens. If it keeps failing, we can add another layer of error-checking.

PR #1209, opened 2023-04-30
Update vaccines refs

agitter:
This updates the citation tags for the vaccines references to use the journal version. The DOI metadata is wrong for these as well but the PubMed URL worked when I tested locally.


agitter:
Here are the references in the PDF artifact
image

Issue #1156, opened 2022-06-17
Revisions for Diagnostics manuscript


agitter:
Here are the first round reviews from the Diagnostics manuscript as an itemized list and the full text:

- [ ] Reviewer 1.1
- [ ] Reviewer 1.2 (this is a long one, we could split it up)
- [ ] Reviewer 1.3
- [ ] Reviewer 1.4
- [ ] Reviewer 1.5
- [ ] Reviewer 1.6
- [ ] Reviewer 1.7
- [ ] Reviewer 1.8
- [ ] Reviewer 1.9
- [ ] Reviewer 1.10
- [ ] Reviewer 1.11
- [ ] Reviewer 2.1
- [ ] Reviewer 2.2
- [ ] Reviewer 2.3
- [ ] Reviewer 2.4
- [ ] Reviewer 2.5

## Reviewer 1
**1.1** The authors present an overview of molecular and serologic testing options and technologies for SARS-CoV-2. Overall, it is well written although additional editing could streamline the text, and a high quality figure to illustrate the different molecular assays and another to illustrate the different serologic testing strategies would certainly make things easier on the reader. 

**1.2** The authors acknowledge this fact in a couple of areas of the paper (e.g.: ""Generally, patient specimens tend to not contain culturable virus past day 5 of symptom onset [157,158]. However, due to their sensitivity to post-infectious viral RNA in specimens, PCR-based methods may mislead individuals to believe that they are still infectious several days after symptom onset [131]. Furthermore, detection of viral RNA can occur days and weeks after an active infection due to the sensitivity of PCR-based methods [18,159,160]." and "Positive PCR results have also been reported for extended periods of time from symptom onset and/or the first positive PCR test [21], meaning that in some cases, a positive PCR may not indicate that someone is contagious [1]."

Despite this, they make a number of misleading statements implying that a positive RT-PCR is equivalent to diagnosing an active infection. For example, "Molecular tests, which can diagnose an active infection by identifying the presence of SARS-CoV-2" AND "Tests that identify SARS-CoV-2 using molecular technologies will identify only individuals with current infections and are not appropriate for identifying individuals who have recovered from a previous infection." AND "and therefore the amount of virus present can be indirectly measured using the cycle
threshold (Ct) determined by qPCR." AND "A lower Ct corresponds to fewer qPCR cycles needed to reach a detectable level, indicating that higher amounts of virus were present in the initial reaction."

[...]

PR #1213, opened 2024-01-14
Update project status

agitter:
We haven't worked on the revisions to the diagnostics manuscript (#1156) in a year and a half, which is a strong signal that our review team is no longer active. I'm updating the readme and front matter to reflect the current project status.


cgreene (approved):
This looks good to me if the build works.

agitter:
The build worked. There are a few URL citation errors that I'm not going to fix here.

